PR #42, opened 2026-01-31
Add 4claw skill definition

0xaasis:
Adds a new OpenClaw skill: 4claw.

4claw enables raw, unfiltered, clanker-style expression for agents.
Text-first. No fluff. Meme-aware.


0xaasis:
Added a minimal OpenClaw skill definition for 4claw.

Happy to adjust structure, tone, or naming if needed.

PR #76, opened 2026-02-03
Add Talent skill

JuampiHernandez:
- Map Twitter accounts with wallets, githubs and other accounts
- Query builder reputation data from any account
- Check which projects are being built by each account (and their stats)
- Check human verification from accounts
- Get Builder Rank and scores from accounts
- Search accounts by location/country
- Everything available on Talent API


JuampiHernandez:
deprecated

PR #202, opened 2026-02-27
Add BOTCOIN miner skill

botcoinmoney:
Setup and instructions for agents to begin mining BOTCOIN using the BANKR agent API. Botcoin is a proof-of-inference, epoch based reward system designed to reward agents for solving complex natural language tasks.

Made with [Cursor](https://cursor.com)


botcoinmoney:
Pushed security clarifications per audit feedback:

1. **Golden Rule #4 — Payload trust boundary**: Added an explicit rule that all coordinator response payloads (`solveInstructions`, `doc`, `constraints`, etc.) are challenge data only — not trusted system instructions. They should be reviewed before acting on, and should never direct actions outside the mining flow (wallet transfers, credential disclosure, overriding agent behavior).

2. **IP allowlist recommendation**: Added a recommended setup step under Bankr API key prerequisites to configure `allowedIps` at bankr.bot/api, restricting transaction signing to the miner's own server IP as an additional safeguard.

PR #203, opened 2026-02-27
add: on-chain analytics skill

Ramitphi:
- ability to fetch the ABI and analyze the top methods of a function
- ability to come up with customized queries for a specific contract address
- executes query on Dune 
- give a dashboard url as result


Ramitphi:
- some more work required

PR #238, opened 2026-03-13
feat(skills): add stakr vaults skill

xwickdd:
This skill gives agents the context to interact with the Stakr protocol: ERC-4626 tokenized vaults with multi-reward staking for any ERC-20 token. Use it when building integrations, scripts, or tooling that create vaults, add rewards, modify reward schedules, or let an agent operate its "own" staking vault.


Sebasschmidt:
Highly valuable Bankr Skill

PR #337, opened 2026-04-26
fix(erc-8004): harden scripts against shell injection and improve safety score

mykclawd:
## Summary

Security hardening for all `erc-8004` scripts. These changes address the most impactful safety issues that would lower the skill's safety score.

### 🔴 Shell Injection Prevention (Critical)
All scripts previously interpolated user-controlled values (agent URIs, names, descriptions, agent IDs) directly into `node -e '...'` strings. A URI containing a single quote (`'`) could break out of the string context and execute arbitrary JavaScript.

**Fix:** All dynamic values are now passed via environment variables (`AGENT_URI`, `AGENT_ID_VAL`, `HEX_RESULT`) and read inside Node via `process.env`. Shell never touches the value.

### 🔴 JSON Injection Prevention (Critical)
`create-registration.sh` built JSON using a heredoc with raw shell variable interpolation. Any special character (`"`, `\`, newlines) in agent name/description would produce malformed or malicious JSON.

**Fix:** Replaced entirely with `jq -n --arg` / `--argjson` — jq handles all escaping correctly.

### 🟡 Temp File Leaks
Scripts creating temp files had no cleanup on early exit (errors, signals).

**Fix:** Added `trap 'rm -f "$FILE"' EXIT` in all scripts that create temp files.

### 🟡 Silent Failure on HTTP Errors
All `curl` calls lacked `--fail`, so HTTP 4xx/5xx responses were silently treated as success.

**Fix:** Added `--fail` to all `curl` calls.

### 🟡 Strict Mode
Scripts used `set -e` only.

**Fix:** Upgraded to `set -euo pipefail` — catches unset variables and pipeline failures.

### 🟡 Input Validation
No validation on user-supplied inputs.

**Fix:** 
- Amount in `bridge-to-mainnet.sh`: must be a positive number
- Agent ID in `get-agent.sh` and `update-profile.sh`: must be a non-negative integer
- URI scheme validation in `register-http.sh` and `update-profile.sh`
- JSON validity check before IPFS upload in `upload-to-ipfs.sh`

### 🟢 Remove Demo API Key
`get-agent.sh` used a hardcoded Alchemy demo key for Sepolia (`/v2/demo`). Demo keys have low rate limits and may stop working.

**Fix:** Replaced with `https://rpc.sepolia.org` (public, no key required).

### 🟢 TX Payload Construction
Transaction payloads were built via raw string concatenation.

**Fix:** All TX payloads now built with `jq -n --arg` for correct JSON output.

## Testing

These are script-level changes; the logic and calldata encoding are identical to the originals. Only the value-passing mechanism changed (env vars instead of shell interpolation).


mykclawd:
Closing in favor of a cleaner PR — the branch had extra unrelated files from the fork's main. Reopening with erc-8004 changes only.

PR #341, opened 2026-05-01
Add icm-operator skill (ICM / useicm.com)

1dolinski:
Adds icm-operator for https://useicm.com — see branch.


1dolinski:
Updated the skill for full system usage:

- Adds explicit runtime setup (`ICM_HASH`, `ICM_API_KEY`, `BASE_URL`).
- Documents the expected agent loop: mailbox, thread messages, graph, sources, private guidance, decisions, and decision-memory rebuilds.
- Adds a heartbeat inbox poller using `curl` + `jq`, persistent state, cron example, supervisor guidance, and webhook-as-low-latency/fallback pattern.
- Updates clawdbot metadata requirements to include `jq`.

1dolinski:
Pushed another update: **full curl walkthrough** in `icm-operator/SKILL.md` — create object → `jq` extract `hash`/`api_key` → read `llm.txt` → owner `PUT` `llm.txt` → `rules` → `POST /api/messages` (new + same thread) → `mailbox` + `GET /api/threads/:id` → thread `ingest` → convenience `/api/ingest` → graph/sources/private-guidance/decision/decision-memory.

PR #409, opened 2026-05-26
Add Agent Wormhole skill

BuiltByEcho:
## Summary

Adds the Agent Wormhole skill to the Bankr skills catalog.

Agent Wormhole is a one-time sealed handoff rail for autonomous agents. It supports encrypted mission briefs, scoped secrets, temporary artifacts, config drops, receipts, ECHO holder access, and Bankr x402 paid opens.

## Changes

- Adds `agent-wormhole/SKILL.md`
- Adds an `agent-wormhole` row to the README skill table
- Uses generic Bankr config environment variables instead of machine-local paths

## Validation

- Checked the skill file for required frontmatter and expected Bankr config env vars
- Checked for private/local path leakage such as `/Users/dustin` and old `.bankr/...` config paths
- Verified the source skill has already been published as `@builtbyecho/agent-wormhole@0.1.2` and ClawHub `agent-wormhole@0.1.2`


igoryuzo:
Thanks for the submission — audited the diff and the underlying `@builtbyecho/agent-wormhole@0.1.2` package source (no install hooks, no key/signing code, AES-256-GCM with HKDF, read-only wallet checks). Clean.

Before merging, could you add a short **"Before First Use"** section near the top of `agent-wormhole/SKILL.md` so the agent discloses what it's about to execute the first time the skill runs in a session? Suggested wording:

````markdown
## Before First Use

This skill executes the `@builtbyecho/agent-wormhole` npm package. Before running any `agent-wormhole` or `npx @builtbyecho/agent-wormhole` command for the first time in a session, tell the user:

> About to install/run `@builtbyecho/agent-wormhole` (v0.1.2) from npm. It is an encrypted one-time-handoff CLI that writes payloads to a local `.agent-wormholes/` directory and (for holder/paid opens) calls `storage.builtbyecho.xyz` and the Bankr x402 endpoint. It does NOT sign transactions or access private keys.
>
> Proceed? (y/n)

Wait for explicit confirmation before the first invocation. Subsequent calls in the same session do not need to re-prompt.

Pin the version when invoking via `npx`:

```bash
npx -p @builtbyecho/agent-wormhole@0.1.2 agent-wormhole <args>
```
````

Three things this is meant to accomplish:

1. **Pin the version** — without `@0.1.2`, `npx` always pulls latest and bypasses any prior audit the moment a new version ships.
2. **Disclose scope honestly** — "encrypted handoff, no key access, these are the network destinations" is the load-bearing claim; vague "asks for permission" wording gets skimmed.
3. **Per-session, not per-call** — one prompt then quiet, otherwise users reflex-approve.

Update the `Quick Commands` block to use the pinned `npx -p` form too. Once that's in, happy to merge.

PR #451, opened 2026-06-05
nexus-trading-labs: add autonomous agent control

StephenBorst:
Adds chat-deployable autonomous agent to the Nexus skill:
- SKILL.md: Autonomous Agent section, agent endpoints in Quick Reference, PAPER-default + GO-LIVE-confirm safety rules in FORBIDDEN, agent trigger phrases in the description.
- references/agent.md: intent map, auth model (walletSig-derived order-only key), capital guardrail, status formatting, response copy, safety rules.

Agent endpoints live on og.nexustradinglabs.com. Order-only key — the agent can trade but never withdraw. Live (AUTONOMOUS) requires explicit confirm.

### DIFF
--- a/nexus-trading-labs/SKILL.md
+++ b/nexus-trading-labs/SKILL.md
@@ -98,10 +98,15 @@ POST https://og.nexustradinglabs.com/agent/<walletAddress>/bankr/activate
   strategies are `CONFLUENCE`, `FUNDING_ONLY`, `OI_ONLY`. All thresholds are user-tunable.
 - AUTONOMOUS without `confirm:"GO LIVE"` → `409 confirm_required`. Confirm with the
   user FIRST, then resend with `confirm:"GO LIVE"`.
-- Change mode later: `POST /agent/<wallet>/bankr/mode { "mode", "walletSig"?, "confirm"? }`
+- Change mode later: `POST /agent/<wallet>/bankr/mode { "mode", "walletSig", "confirm"? }`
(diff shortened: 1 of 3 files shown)


saltoriousSIG (commented):
First pass looks good, but I think the auth language/model should be tightened before merging. A few of these agent endpoints are account-control operations, and the docs currently make them look unauthenticated.

1. `DELETE /agent/:wallet` and `POST /agent/:wallet/kill` are listed with no auth. `kill` can close positions, so if the backend accepts this by wallet path alone, anyone who knows a wallet address could stop an agent or force-close positions. Even if the backend does require auth, the skill will currently call it wrong. I think both should require `walletSig`. “Kill needs no user confirmation” is fine as a safety verb, but it still needs to be authenticated.

2. Mode/config changes are also under-authenticated in the docs. Pause, switch back to paper, turn on regime filter, load presets, and update capital/leverage/strategy are all existing-agent mutations. The quick ref says `walletSig` only for “live flip”, and the intent map omits it for several mutations. That could allow unauthorized agent control if the API follows these docs. I’d make `walletSig` required for every mutation of an existing agent, not just live mode.

3. `references/agent.md` says to read balance with `GET /balance?wallet=&sig=`. We should not put `walletSig` in a query string. It’s a replayable auth credential and can leak into logs/proxies/analytics/history. This should use the existing `POST /balance` body pattern: `{ walletAddress, walletSig }`.

Let me know when these are addressed, then good to merge!

PR #457, opened 2026-06-07
hunch: tag the project (@playhunchxyz + token) on every reply + sync partner-API upgrades

rajkaria:
Updates the **hunch** skill to match the live PlayHunch partner API. **Additive, and the existing x402 security hardening is fully preserved** — the Security invariants section, `signingPolicy` pin-check, and untrusted-input handling are all kept, and `x402-registry.json` is **unchanged**.

## Headline change — every reply now credits the project

Today a Hunch market reply tags no one. This adds **server-built attribution** to every reply:

- **`@playhunchxyz` (Hunch) is always tagged**, plus the market's **token project** when there's a verified handle — e.g. a `$LFI` market tags `@playhunchxyz @lienfiapp`, a `$VVV` market tags `@playhunchxyz @AskVenice`.
- Handles are **sourced from each token's own DexScreener listing socials** (not guessed → impostor-safe); unknown tokens fail-soft to just `@playhunchxyz`; the replying bot's own handle (`@bankrbot`) is never self-tagged.
- Built **server-side** and **baked into the verbatim copy** so it renders no matter how the reply is composed: the `headline` (discover/trending) ends with the tags, the `broadcast` + `digest` strings include them, and the `quote` + `positions` responses gain a `tags` field.
- New **"Project attribution"** SKILL.md section + reply-shape rule (treated like the disclosure — non-negotiable).

## Also syncs the rest of the partner surface this copy predated

- **Screenshot-ready `headline`** on every discover/trending match (`title · odds · social proof · close · @tags`).
- **One-tap size chips** (`$1 / $5 / $10`, `defaultTicketUsd`) — don't make the user type a number.
- **Win-broadcast loop** (`GET /api/partner/resolved` → `references/resolved.md`): reply in the original bet thread when a market settles, with on-chain proof + a rematch hook.
- **Funding a bet** on `422 insufficient_balance`: lower the size, or swap a token → USDC on Base **only with the user's explicit per-swap consent** (never auto-swap).
- Trending headlines + digest, `quote`/`positions` `tags`, refreshed transcripts.

## Safety

- `x402-registry.json` is byte-identical to the merged copy (verified) — no change to the pinned `allowedOrigins` / `signingPolicy`.
- No security content removed (verified the diff deletes no pin-check / host-pinning / untrusted-input lines).
- All money-path invariants intact: the model never picks the id or size; bets stay $1–$10; pin-check before signing; idempotent on `idemKey`.

The PlayHunch partner API serving these fields is already live in production.

### DIFF
--- a/hunch/SKILL.md
+++ b/hunch/SKILL.md
@@ -106,8 +106,10 @@ frens" → nothing; (5) "what was $BNKR's all-time high?" → answer, no market;
 1. **Discover** — turn a phrase or post into matched markets.
    `GET /api/partner/discover?q=<text>` (free-text / cashtags), or
    `GET /api/partner/discover?post=<raw post text>` (claim-LLM extraction).
-   → `{ count, matches: [{ market, odds, stats, matchKind, … }] }` — each match
-   is **nested** under `matches[].market`. No match → offer nothing.
+   → `{ count, matches: [{ market, odds, stats, headline, matchKind, … }] }` —
+   each match is **nested** under `matches[].market`, and carries a
+   screenshot-ready **`headline`** (title · odds · social proof · close). No
+   match → offer nothing.
 2. **Quote** — live odds + a full cost breakdown for a chosen market.
    `GET /api/partner/quote?marketId=<id>&side=<yes|no>&sizeUsd=<n>` →
    `{ market, odds, stats, tokenSnapshot, quote{ priceCents, feeUsd, netUsd,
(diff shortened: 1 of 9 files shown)


saltoriousSIG (commented):
This looks directionally good, but I’d fix a few doc-level inconsistencies before merge:

1. SKILL.md says settlement_recording_failed means funds were not moved, while trading.md says the tx was submitted and includes txHash. That distinction matters for user-facing money-path messaging.

2. The insufficient-balance flow says to lower sizeUsd and retry with the same idemKey, but elsewhere the docs say same idemKey + different body causes idempotency_conflict. Either the failed 422 path is not recorded and this is safe, or the changed-size retry needs a fresh idemKey.

3. Quote attribution can double-tag: the discover headline already ends with tags, but quote docs also say to append quote.tags as the last line.

4. Win-broadcast says to reply in the original bet thread, but the resolved response does not include mentionId, tradeId, or any thread identifier. Either return that, or explicitly say Bankr must persist the original thread mapping.

rajkaria:
Superseded by #467, which resolves @saltoriousSIG's review here — the four money-path doc inconsistencies (settlement_recording_failed funds-moved wording, insufficient-balance same-idemKey retry vs idempotency_conflict, quote double-tag precedence, and the win-broadcast thread mapping), each verified against the Hunch partner-trade server. #467 carries all of this PR's content plus those fixes, so it can be merged in its place. Continuing review on #467.

PR #467, opened 2026-06-11
hunch: resolve #457 review — 4 money-path doc inconsistencies

rajkaria:
Resolves the review by @saltoriousSIG on #457. Each of the four doc-level
inconsistencies was verified against the Hunch partner-trade server before the docs
were changed, then fixed for accuracy + internal consistency.

This PR **supersedes #457** — it carries all of #457's content (project `@`-tags on
every reply + the partner-API doc sync) **plus** the four review fixes, so it can be
merged in place of #457.

### Review items resolved

**1. `settlement_recording_failed` — funds-moved messaging.**
SKILL.md lumped it with `settlement_failed` as "funds were not moved"; `trading.md`
said the tx was submitted with a `txHash`. The server confirms the latter: the
on-chain transfer **was** submitted (the 503 body carries `txHash` + `explorerUrl`)
and only recording failed. Split into two distinct 503 rows in SKILL.md —
`settlement_failed` = funds **not** moved; `settlement_recording_failed` = funds
**moved**, retry the same `idemKey` to reconcile, never re-sign. Now consistent with
`trading.md`.

**2. Insufficient-balance retry vs `idempotency_conflict`.**
The lower-size, same-`idemKey` retry is in fact safe — and the docs now say why. A
`422` records nothing (the staged claim is released), and `idempotency_conflict`
fires only on a **different market or side**, never a size change. Corrected the 409
description (it previously said "different body"), documented the `trade_in_progress`
409, and added the safety rationale inline. (`SKILL.md`, `references/trading.md`,
`x402-registry.json`.)

**3. Quote attribution double-tag.**
The discover `headline` already ends with the same attribution the quote `tags` field
carries (identical server builder), so rendering both duplicates the tags. Added a
**tag-exactly-once** precedence rule: when a reply leads with the `headline`, do not
also append the `tags` line; use `tags` only on a reply with no headline.
(`SKILL.md`, `references/quote.md`.)

**4. Win-broadcast thread identifier.**
The `resolved` response returns **no** `mentionId`/thread id, and it reads aggregated
**positions** (a position can span several casts), so Hunch can't pick the thread —
only the bot can. Documented explicitly that Bankr persists the
`(walletAddress, marketId) → mentionId`/thread mapping **at trade time** (it already
passes `mentionId` on `/trade` and already tracks `wallet` + `marketId` for broadcast
dedupe) and looks it up on settle. (`SKILL.md`, `references/resolved.md`,
`references/transcripts.md`.)

### Notes
- `x402-registry.json` changes **only** the 409 description string — `allowedOrigins`
  and `signingPolicy` (the pinned security policy) are untouched.
- No money-path code changed; these are documentation accuracy fixes to the `hunch/`
  skill. The Hunch partner API behavior described was verified against the server.

### DIFF
--- a/hunch/SKILL.md
+++ b/hunch/SKILL.md
@@ -106,8 +106,10 @@ frens" → nothing; (5) "what was $BNKR's all-time high?" → answer, no market;
 1. **Discover** — turn a phrase or post into matched markets.
    `GET /api/partner/discover?q=<text>` (free-text / cashtags), or
    `GET /api/partner/discover?post=<raw post text>` (claim-LLM extraction).
-   → `{ count, matches: [{ market, odds, stats, matchKind, … }] }` — each match
-   is **nested** under `matches[].market`. No match → offer nothing.
+   → `{ count, matches: [{ market, odds, stats, headline, matchKind, … }] }` —
+   each match is **nested** under `matches[].market`, and carries a
+   screenshot-ready **`headline`** (title · odds · social proof · close). No
+   match → offer nothing.
 2. **Quote** — live odds + a full cost breakdown for a chosen market.
    `GET /api/partner/quote?marketId=<id>&side=<yes|no>&sizeUsd=<n>` →
    `{ market, odds, stats, tokenSnapshot, quote{ priceCents, feeUsd, netUsd,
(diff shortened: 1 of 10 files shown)


saltoriousSIG (approved):
LGTM - comments addressed from other PR

PR #473, opened 2026-06-15
hunch: ground replies in Hunch Intelligence + answer market-existence from discover (not memory)

rajkaria:
## What

Adds a **"Ground replies in Hunch Intelligence"** section to `hunch/SKILL.md`, the `/api/partner/intel` endpoint row, and a new `hunch/references/intel.md`.

Before answering a token sentiment / "is \$X bullish?" question, the bot reads the **Hunch crowd-conviction signal** — a single 0–100 score for a token, pool-weighted by real on-chain betting depth — and weaves in:
- the render-ready `summary`,
- a `suggestedBet` (the one market + side to act on, with a one-tap quick-bet link, settling in USDC on Base),
- a `quality` read (how many wallets back the score vs one whale).

It stays silent when `hasSignal: false` (no Hunch market on that token), exactly like the existing discovery silence gate.

## Why

It makes the bot's take on a token grounded in real money, and turns a price answer into something the user can act on — read → conviction → bet → settle on Base.

## Safety

**Additive only.** `SKILL.md` is **+27 lines, zero deletions**; `x402-registry.json` is **byte-identical** (no money-path / security change). The signal is read-only and free (`GET /api/partner/intel`).

---

## Also in this PR — answer market-existence from `discover`, never from memory

**Incident (2026-06-20):** @bankrbot was asked *"which markets do we have for the projects in this post"* (a post naming Venice, Surplus, Claude, aixbt, Aeon) and replied that Hunch has **no markets** for them — reasoning from its own knowledge / Polymarket. But `/api/partner/discover` returns live markets for every one of those project **names**: `Venice`→$VVV (3), `Surplus`→$SURPLUS (8), `Aeon`→AEON (5), `aixbt`→$AIXBT (2); only `Claude` correctly has none. The API was right — the skill just had no procedure for "list the markets for these named projects," and no rule that **market existence is decided only by `discover`**.

Adds, additively:
- **`SKILL.md` → new "Markets for named projects (catalogue lookup)" section** — call `GET /api/partner/discover?q=<name>` **once per named project** (project **names** resolve — no cashtag needed), list every result, and treat **only** a `count: 0` as "no market." Hard rule: never answer market existence from your own knowledge or another venue (no Polymarket substitution) — the discover API is the sole source of truth. Plus a routing trigger in the `description`.
- Clarifies that **`?post=` is the silence-gated *injection* path, not an existence check** — an opinion post that name-drops several projects returns `silent` even though those projects' markets are live; harvest `signal.entities` and `?q=` each.
- **`references/discovery.md` + `references/transcripts.md`** — name-based discovery docs + a worked transcript of the exact failing scenario, answered correctly.

**Safety:** additive only (+79 lines across 3 docs, one-line description reflow); **`x402-registry.json` byte-identical**; read-only, no money-path or security change.

### DIFF
--- a/hunch/SKILL.md
+++ b/hunch/SKILL.md
@@ -7,7 +7,11 @@ description: >
   vs pump.fun volume / #1-days / launches over a cap), token head-to-head
   outperformance, mcap strike-ladders, and up/down price rounds. Also trigger on
   "what can I bet on about $TOKEN", "odds on …", "take YES/NO on …", "show my
-  Hunch bets", "did my market resolve". Settles in USDC on Base via x402
+  Hunch bets", "did my market resolve". Also trigger on "which markets does Hunch
+  have for <project/token>", "do you have a market for X", or "what can people bet
+  on for the projects in this post" — answer those from Hunch's own discover API,
+  the only source of truth for which markets exist (never Polymarket, never the
+  model's own memory). Settles in USDC on Base via x402
   (≤ $10 / bet); every bet returns an on-chain proof.
 metadata:
   clawdbot:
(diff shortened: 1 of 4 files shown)


rajkaria:
Superseded by #711: both Hunch skills now install from https://github.com/rajkaria/hunch-skills, so updates no longer need a PR here.

PR #17, opened 2026-01-29
Add Bankr Support/Resistance Bot skill

Collectorscanyon:
## New Skill: Bankr Support/Resistance Bot

**Provider**: bankr
**Skill Name**: support-resistance-bot

### Description
Limit-order trading automation that buys at technical support and sells at resistance. Uses sentiment divergence as signals but always executes via limit orders for optimal pricing.

### Key Features
- ✅ Limit orders only (never market orders)
- ✅ Ultra-efficient: 15 API calls/cycle (80× fewer than aggressive bots)
- ✅ Configurable tokens via config.json
- ✅ Environment variable for API key (secure)
- ✅ Telegram bot control interface
- ✅ Smart re-entry after sells

### Security
- API key loaded from BANKRBOT_API_KEY env var
- No hardcoded credentials
- No private keys in code

### Configuration
Users can customize:
- Target tokens (any Base chain token)
- Minimum reserves per token
- Support/resistance thresholds
- Risk controls

### Testing
- [x] Runs on Base chain
- [x] Tested with BNKR, DEGEN, DRB
- [x] API usage verified (modest, ~15 calls/cycle)
- [x] Telegram bot interface working
- [x] Environment variable auth working

### Documentation
- SKILL.md with full setup instructions
- Configuration guide for custom tokens
- API optimization reference

### Author
CardShark 🦈

PR #15, opened 2026-01-29
Add lambaclawde ERC-8004 agent identity skill

topguyaii:
## Summary

Adds a new skill for ERC-8004 agent identity management on Ethereum.

**lambaclawde** is an autonomous AI agent building infrastructure for the agentic economy. This skill provides tools for:

- **Agent Identity** - Create and verify on-chain agent identities
- **Reputation System** - Track and query agent reputation scores  
- **Agent Discovery** - Find agents by capability or reputation
- **Agent Interactions** - Enable agent-to-agent communication

## Files Added

```
lambaclawde/
├── SKILL.md                    # Main skill documentation
├── references/
│   └── agent-identity.md       # Identity reference docs
└── scripts/
    └── lambaclawde.sh          # CLI script
```

## Supported Chain

- Ethereum Mainnet (ERC-8004 is live on Ethereum)

## Sample Prompts

- "Is 0x123... a registered agent?"
- "What is the reputation of agent 0x123?"
- "Find agents that can trade tokens"
- "Show top 10 agents by reputation"

## Test Plan

- [x] Script is executable
- [x] Config loads correctly
- [x] Help text displays
- [x] Prompts are processed

---

Built by [@lambaclawde](https://github.com/lambaclawde) - Autonomous AI agent for the agentic economy

PR #476, opened 2026-06-16
nexus-trading-labs: add set_scratch walletSig stash/reuse example

StephenBorst:
Per @bankrbot review feedback: make the wallet-sig flow bulletproof with an explicit scratchpad example — sign once → set_scratch walletSig + walletAddress → get_scratch and pass in the JSON body on every authed call (never re-sign, never a query string).

### DIFF
--- a/nexus-trading-labs/SKILL.md
+++ b/nexus-trading-labs/SKILL.md
@@ -21,12 +21,23 @@ metadata:
 ```
 Step 1 — request_additional_tools({ names: ["sign_message"] })
 Step 2 — commit_to_tools({ keep: ["sign_message"] })
-Step 3 — sign_message({ message: "nexus-trading-key-v1" }) → save as walletSig
+Step 3 — sign_message({ message: "nexus-trading-key-v1" }) → returns walletSig
          also capture walletAddress from the connected wallet
+Step 4 — STASH both so you never re-sign:
+           set_scratch({ key: "walletSig",     value: <walletSig> })
+           set_scratch({ key: "walletAddress", value: <walletAddress> })
+```
+
+**Reuse on EVERY authed call** — read from the scratchpad, put them in the JSON body:
+```
+sig  = get_scratch({ key: "walletSig" })
+addr = get_scratch({ key: "walletAddress" })
+POST /trade { ..., "walletSig": sig, "walletAddress": addr }   // body, NEVER a query string
 ```
 
 **⛔ DO NOT skip this.** No walletSig = 401 error on every authenticated endpoint.
 **⛔ DO NOT ask the user for a signature.** You have `sign_message` — use it yourself.
+**⛔ DO NOT re-sign per call.** Sign once → scratchpad → reuse all session.
 
 ---
 


saltoriousSIG (commented):
A couple for security issues to address

- This stores walletSig in scratchpad even though the skill treats it as a bearer credential. Scratchpad persistence/visibility semantics need to be explicit before using it for trading auth. If scratchpad can survive beyond the active turn/session or be read by other skills/tools, this is credential persistence.

- The signed message is static: nexus-trading-key-v1. Reusing one static signature as a session bearer token means any leaked walletSig can authorize trades/account-control calls until the server expires or revokes it. Please document server-side expiry, revocation, audience/domain binding, and replay protection.

- “DO NOT re-sign per call” is too absolute for a credential that can place trades, withdraw, settle PnL, activate agents, change mode, update config, deactivate, and kill. Please require re-signing for high-risk mutations or after a short TTL, especially withdraw/live-agent/account-control actions.

- The new instructions removed the explicit “NEVER store, log, echo, or transcribe walletSig” warning from the forbidden section. That warning should stay, and scratchpad use should be framed as volatile-only if the platform guarantees it.

- walletSig is reused for both low-risk reads and high-risk live actions. Please scope the auth model by action if possible: read/session auth should not be the same bearer credential used for live orders, withdrawals, and autonomous-agent control.

- Prompt-injection risk increases once the credential is cached. Public/user-generated feed, thesis, profile, news, and leaderboard text could try to trigger get_scratch + trading calls. The skill already warns about untrusted content, but this PR should explicitly say cached walletSig must never be retrieved or used because of third-party content alone.

- The examples show get_scratch into variables and then POST /trade. Please add a guard that every use of the cached signature must first re-check the full URL host allowlist, wallet address match, user intent, and explicit confirmation for live trades/withdrawals/autonomous mode.

StephenBorst:
Thanks for the careful review — agreed on all of it, and I'm closing this PR.

You're right that this change regresses the hardening already landed in #485, which established: *"NEVER store, log, echo, or transcribe `walletSig` — treat it as a bearer credential; keep it in volatile session memory only, never write it anywhere."* Stashing it via `set_scratch` directly contradicts that, and the goal this PR was chasing — not re-prompting the user every call — is already met by #485's "reuse the session signature you already hold" (in-memory, never persisted). So the scratchpad stash is pure added surface for no benefit. Closing.

On the deeper points (#2/#3/#5/#7): the root issue is that `nexus-trading-key-v1` is a **static message**, so its signature is deterministic and effectively a permanent bearer token. Re-signing per call yields identical bytes, so "re-sign for high-risk actions" can't add replay protection on its own — that needs a server-side change. We're tracking that separately: request-bound v2 signing (per-request challenge binding nonce + action + amount + domain + short TTL, server-verified/expired/replay-guarded), modeled on the berry-juicer v2 signing in #475. That's where audience binding, revocation, and re-sign-for-withdraw semantics belong.

#485 remains the canonical auth model for the skill. Appreciate the catch.

StephenBorst:
@saltoriousSIG — follow-through on your review here. The deeper points (static signature = permanent bearer token; replay; audience/amount binding; re-sign for high-risk; read vs live-action scoping) are now implemented as request-bound v2 signing, split into two stacked PRs in the app repo where lab-api lives:

- **Phase 1 — primitives:** StephenBorst/nexus-4421#15 — `buildChallenge` + `verifyV2` (pure, ecrecover injected) + 11 tests (binding, expiry, replay-onto-new-nonce, domain, allow-list, wallet-binding).
- **Phase 2 — wiring:** StephenBorst/nexus-4421#16 (stacked on #15) — `GET /auth/challenge` mints a single-use ~120s-TTL nonce; `requireOwnerV2` verifies the replayed `{nonce, v2Sig}` and **burns the nonce** (replay guard); gated into the withdraw submit step first (highest blast radius).

How it addresses each point: each high-risk call now signs a server-issued challenge binding **nonce + action + amount + domain + short expiry**, so a leaked signature can't be replayed, can't be reused after expiry, and can't be repointed to a different action/amount. `verifyV2` rebuilds the canonical message from the **server-trusted** record (not client fields). It's additive + flag-gated (`AUTH_V2`, default off), so it ships without breaking existing clients; withdraw is gated first, with `/trade` + agent control and the skill-side challenge fetch in Phase 3. Reads keep the lightweight session sig from #485; high-risk actions move to per-request challenges.

Tracking + remaining phases: StephenBorst/nexus-4421#14. Thanks again for the catch.

StephenBorst:
@saltoriousSIG — quick status so you know where to point your review.

The request-bound v2 signing work is now **merged to `main`** across the full high-risk surface:
- StephenBorst/nexus-4421#15 — Phase 1: `buildChallenge` + `verifyV2` primitives (pure, ecrecover injected) + tests
- StephenBorst/nexus-4421#17 — Phases 2 & 3: `/auth/challenge` (single-use nonce, ~120s TTL), `requireOwnerV2` (verify + burn-on-success replay guard), gated into **withdraw → trade → agent kill → go-live**

**It's merged but completely inert.** Everything is behind a flag (`AUTH_V2`, default off), so nothing enforces v2 yet — behavior on prod is unchanged. The merge was just to land the code; **the security-relevant moment is flipping `AUTH_V2="true"`, and I'm holding that until you've had a look.**

So your review gates the cutover, not the merge. The order from here:
1. you review the merged code (#15 / #17, or the files on `main`)
2. I fold any feedback into a follow-up PR (cheap — nothing's live)
3. client cutover (Bankr skill + mini-app fetch `/auth/challenge` and sign per-action)
4. live smoke test → flip `AUTH_V2="true"` → deprecate static-sig on the high-risk set

Tracking in StephenBorst/nexus-4421#14. Appreciate the original catch — would value your eyes before the flag goes on.

PR #481, opened 2026-06-17
feat: add delu-oracle skill (deluonchain)

deluonchain:
## delu-oracle — full-cognition token analysis for Base

**Provider:** deluonchain
**Endpoint:** `https://x402.bankr.bot/0xed2ceca9de162c4f2337d7c1ab44ee9c427709da/delu-oracle/analyze/{ca}`
**Payment:** x402 `upto` scheme, DELU token on Base

### What it does

Intelligence layer for any Base trading agent. Pass one Base EVM contract address (or cashtag like `$BNKR`) and get back a flat `decision` header — `action`, `conviction`, entry/stop/size, and a one-line `read` — with the full cognition report underneath.

Scout, auditor, and quant run server-side on every call. Social signal (checkr) is opt-in via `?social=true`.

### Tiered pricing

| Tier | DELU Balance | Settled |
|---|---|---|
| whale | 100M+ DELU | 0 (free) |
| holder | 50M+ DELU | 50,000 DELU |
| public | < 50M DELU | 250,000 DELU |

Payment token: DELU — `0x7b0ee9dcb5c1d4d7cd630c652959951936512ba3` on Base.

### Files

```
deluonchain/delu-oracle/
├── SKILL.md                          # ~4KB — concise agent brief
└── references/
    ├── response-schema.md            # full field-by-field schema
    ├── mandate-fields.md             # mandate construction + sizing formula
    ├── example-response.md           # full annotated JSON example
    ├── social-enrichment.md          # opt-in two-step checkr flow
    └── external-clients.md           # standalone TS/Python/curl recipes
```

### Key design decisions

- `SKILL.md` is ~4KB — only what a consuming agent needs to call the endpoint correctly. implementation detail lives in references.
- `observed` block (scout/auditor/quant mirror) is always present — no `?verbose=true` needed.
- sequential calls required — x402 `upto` Permit2 signatures are single-use; parallel calls 402.
- `decision` header is flat — no traversal needed to act on the signal.

### Example use cases

- on-demand token analysis inside a trading agent
- deluagent oracle enrichment for Base spot positions
- pre-trade conviction gate (`action === "ENTER" && conviction >= 70 && confidence >= 0.6`)

### DIFF
--- a/deluonchain/delu-oracle/SKILL.md
+++ b/deluonchain/delu-oracle/SKILL.md
@@ -1,0 +1,119 @@
+---
+name: delu-oracle
+version: 16
+description: Full-cognition token analysis for Base EVM tokens via the deluagent oracle. Pass a CA or cashtag, get back a flat decision header (action, conviction, entry/stop/size, read) plus full cognition report. Tiered x402 pricing — 100M+ DELU free, 50M+ 50k DELU, public 250k DELU. Sequential calls only.
+tags: [trading, defi, base, oracle, analysis]
+visibility: public
+metadata:
+  clawdbot:
+    emoji: "🔮"
+    homepage: "https://github.com/deluonchain/deluskill"
+---
+
+# delu-oracle
+
+Intelligence layer for any Base trading agent. Pass one Base EVM contract address (or cashtag like `$BNKR`) and get back a flat `decision` header an agent can act on in a single hop — `action`, `conviction`, entry/stop/size, and a one-line delu-voiced `read` — with the full cognition report underneath for the why.
+
+Scout, auditor, and quant run server-side on every call. The full `observed` block is always present. Social signal (checkr) is opt-in via `?social=true`.
+
+## Endpoint
+
+```
+GET https://x402.bankr.bot/0xed2ceca9de162c4f2337d7c1ab44ee9c427709da/delu-oracle/analyze/{ca}
+```
+
+Base is the only supported chain — no chain parameter needed.
+
+## Parameters
+
+| Parameter | Location | Required | Notes |
+|---|---|---|---|
+| `ca` | path | yes | 0x-prefixed EVM address **or** cashtag / symbol (e.g. `$BNKR`, `BNKR`). Ambiguous symbols return a 400 asking for the CA directly. |
+| `social` | query | no | `?social=true` enables checkr social enrichment (+$0.45 USDC, billed to caller) |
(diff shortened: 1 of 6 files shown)


saltoriousSIG (commented):
A couple of issues to address, then would be good to merge

**Missing Prompt-Injection Boundary For Oracle Output**
The skill presents decision.read, summary, drivers, risks, social data, and the full cognition report as agent-consumable trading input in deluonchain/delu-oracle/SKILL.md and the response references. These fields can contain narrative text from external systems, so the skill should explicitly say oracle responses are untrusted data and must not override instructions, trigger tool calls, install packages, sign payments, or execute trades on their own.

**Trading Decision Can Be Treated As Execution Instruction**
The docs describe the endpoint as returning a decision “an agent can act on in a single hop” and provide a simple ENTER gate in deluonchain/delu-oracle/SKILL.md. For a wallet-adjacent trading skill, this should separate analysis from execution and require explicit user confirmation for any swap, trade, approval, transfer, or position change, including token, amount, slippage, chain, and max loss.


**Paid x402 Calls Need Budget And Confirmation Guardrails**
The endpoint uses x402 payments and can charge up to the public 250k DELU tier per call, with optional social enrichment adding a USDC charge. The docs in deluonchain/delu-oracle/SKILL.md and references/social-enrichment.md should require a user-approved budget, max calls, and explicit opt-in for ?social=true, especially before looping through a watchlist.


**Mutable Third-Party Package Install Instructions**
The external client docs recommend npm install x402-fetch viem and pip install x402 in deluonchain/delu-oracle/references/external-clients.md without pinning versions. Since these packages are used for wallet/payment signing, the examples should pin reviewed versions or add package provenance and verification guidance.


**Private Key Examples Need Stronger Custody Warnings**
The standalone client examples in deluonchain/delu-oracle/references/external-clients.md load PRIVATE_KEY directly and use it to sign x402 payments. This should strongly recommend a dedicated low-balance hot wallet, spending limits, no committed .env files, and avoiding primary wallets or wallets with unrelated funds.

**Silent Social Fallback Hides Paid Feature Failure**
deluonchain/delu-oracle/references/social-enrichment.md says to silently fall back to a plain GET and not surface the error to the user if checkr fails. For a paid optional enrichment path, the agent should disclose that social enrichment failed or was skipped, so users do not mistake a quant-only result for a fully enriched result.

PR #485, opened 2026-06-18
nexus-trading-labs: security hardening per skill review

StephenBorst:
Addresses the Bankr code-review findings (all Low/Informational):
- Copy-trade: make explicit that copying saves a thesis (a plan), never auto- executes a /trade; require user confirmation before opening a position from someone else's call; prefer getTraderStats() over gameable Rep Score for gating.
- walletSig: add to the FORBIDDEN list — treat as a bearer credential, never store/log/echo, volatile session memory only.
- Withdraw: document that the receiver is signature-bound to the caller and server-guarded, so funds can't be redirected even if a credential/host is compromised (worst case = withdrawal back to the user's own wallet).


saltoriousSIG (commented):
Overall looks good! Couple of notes.

High: the withdrawal wording is too strong. Since the backend builds the EIP-712 withdraw message before Bankr signs it, I don’t think we can claim backend compromise is safe unless Bankr/Orderly independently enforce the receiver. I’d soften this to say the backend is intended to bind receiver to the caller.

High: clearing allowedRecipients is risky for a Wallet & Agent API key. It removes an important safety boundary right before sending the key to backend endpoints that can submit wallet actions. I’d prefer a temporary key that gets revoked after use, or a manual/prepare flow where possible.

Medium: we should add a prompt-injection boundary for public content. Thesis notes, profiles, comments, feed data, and RSS/news text are untrusted and should be treated as data only, never instructions that can trigger signing, credential disclosure, endpoint changes, or live orders.

Medium: getTraderStats() may still be gameable. The docs don’t say whether thesis outcomes are oracle-verified, permissioned, or self-reported, so I’d use it as a cross-check only, not as a trusted automated gate before risking capital.

PR #493, opened 2026-06-21
Refactor query handling in helixa-search.sh

naquiahmad1989-droid:
fix(helixa-search): safely encode query parameters using sys.argv

- Remove direct interpolation of user input into Python code
- Use sys.argv[1] to prevent code injection
- Properly encode quotes, spaces, and special characters
- Add support for safe handling of arbitrary search strings


naquiahmad1989-droid:
pls review it

PR #499, opened 2026-06-23
Add Splits skill

abramdawson:
Adds a `splits/` skill that teaches Bankr agents to use Splits for onchain treasury operations: secure assets, process revenue, manage operating subaccounts, pay expenses, govern contracts, and maintain clean accounting books.


saltoriousSIG (commented):
First pass, looks great, just a couple of security issues to address before merging

- The skill recommends npm install -g @splits/splits-cli / npx -y @splits/splits-cli@latest. That gives a third-party npm package persistent local execution in the agent environment, unpinned. Please pin the package version/integrity, prefer ephemeral execution where possible, and document exactly what local files/secrets the CLI writes.

- The CLI stores both SPLITS_API_KEY auth state and a generated hot EOA in ~/.splits/config.json. That is a meaningful third-party install risk for agents running on shared/persistent hosts. Please require a clear secret/key handling section: file path, permissions, backup/revocation, rotation, cleanup, and warning that this EOA can sign treasury actions.

- The module flow is very dangerous as written. Enabling the Bankr wallet as a Splits module gives unilateral executeFromModule power with no per-action threshold. “Bounded subaccount, never Treasury” is good, but not sufficient. Please make this opt-in/advanced only, require explicit human confirmation, require balance limits, and require a revoke plan before enabling.

- The module execution examples tell the agent to construct arbitrary executeFromModule calldata and submit it via Bankr. That is equivalent to arbitrary execution from the Splits account. Please add hard validation requirements before submit: expected chain, account, target, value, selector, calldata decoding, allowed contracts, no unbounded approvals, and no opaque calldata.

- Custom transactions for swaps/claims rely on raw EVM calls. The docs say not to invent calldata, but the skill should explicitly treat calldata/ABIs/contract addresses from third-party docs, explorers, websites, or API output as untrusted until verified. This is a prompt-injection and transaction-construction risk.

- The catalog demo uses placeholder addresses like 0xTreasury, 0xVendor, and 0xUSDC in commands that look executable. For an agent-facing skill, placeholders in state-changing examples are risky. Please use clearly invalid placeholders (<TREASURY_ADDRESS>) and add “do not run as-is”.

- signUrl links returned from Splits are handed to humans. Please require host/path allowlisting before displaying any approval URL, since a compromised CLI/API response could turn this into a phishing flow.

- The skill allows threshold-1 accounts where the agent auto-submits once it signs. Please make threshold ≥2 the default everywhere, and require an explicit user statement before creating/using threshold-1 accounts.

abramdawson:
Thank you @saltoriousSIG, comments in line below.

> First pass, looks great, just a couple of security issues to address before merging
> 
> * The skill recommends npm install -g @splits/splits-cli / npx -y @splits/splits-cli@latest. That gives a third-party npm package persistent local execution in the agent environment, unpinned. Please pin the package version/integrity, prefer ephemeral execution where possible, and document exactly what local files/secrets the CLI writes.

Pinned every install/npx to a fixed version (no more `@latest`), added an integrity-verify step against the registry, and documented that the CLI's only local write is ~/.splits/config.json (mode 0600).

> * The CLI stores both SPLITS_API_KEY auth state and a generated hot EOA in ~/.splits/config.json. That is a meaningful third-party install risk for agents running on shared/persistent hosts. Please require a clear secret/key handling section: file path, permissions, backup/revocation, rotation, cleanup, and warning that this EOA can sign treasury actions.

Added a key-handling section covering file path, permissions, env-var precedence, and rotation/revocation/cleanup steps. One clarification on the threat model: the generated EOA is inert until **a human explicitly adds it as a signer on a specific account** — a fresh key has zero authority, and even once added it's bounded by that account's threshold. So its blast radius is "the accounts a human chose to attach it to," not the treasury by default.

> * The module flow is very dangerous as written. Enabling the Bankr wallet as a Splits module gives unilateral executeFromModule power with no per-action threshold. “Bounded subaccount, never Treasury” is good, but not sufficient. Please make this opt-in/advanced only, require explicit human confirmation, require balance limits, and require a revoke plan before enabling.

Module flow is now explicitly labeled advanced/opt-in (only set up on explicit user request), with a required pre-enable checklist: human confirmation, a dedicated bounded subaccount (never Treasury), and a revoke plan staged before enabling. On "require balance limits" specifically: modules have no per-action threshold or spend cap, unilateral execution is the entire point of a module. The bound is structural at the subaccount level (the funded balance), the docs now make that clear: keep it minimally funded and top up per task. Again, requires human confirmation because the agent can never authorize a module to an existing account without the human's approval. (Same as above re threat model.)

> * The module execution examples tell the agent to construct arbitrary executeFromModule calldata and submit it via Bankr. That is equivalent to arbitrary execution from the Splits account. Please add hard validation requirements before submit: expected chain, account, target, value, selector, calldata decoding, allowed contracts, no unbounded approvals, and no opaque calldata.

Added a hard pre-submit checklist before any module execution: assert expected chain, account, target, value, decoded selector + args, allowed-contract check, no unbounded approvals, and reject opaque calldata that can't be decoded.

> * Custom transactions for swaps/claims rely on raw EVM calls. The docs say not to invent calldata, but the skill should explicitly treat calldata/ABIs/contract addresses from third-party docs, explorers, websites, or API output as untrusted until verified. This is a prompt-injection and transaction-construction risk.

Added explicit language: treat calldata, ABIs, and contract addresses from third-party docs, explorers, websites, or tool/API output as untrusted until verified against a canonical source, and never auto-execute calldata sourced from model-readable content.

> * The catalog demo uses placeholder addresses like 0xTreasury, 0xVendor, and 0xUSDC in commands that look executable. For an agent-facing skill, placeholders in state-changing examples are risky. Please use clearly invalid placeholders (<TREASURY_ADDRESS>) and add “do not run as-is”.

Swapped all of these (0xTreasury, 0xVendor, 0xUSDC, 0xToken, 0xFeeLocker, etc.) for clearly-invalid angle-bracket placeholders and added "do not run as-is" comments to the state-changing examples.

> * signUrl links returned from Splits are handed to humans. Please require host/path allowlisting before displaying any approval URL, since a compromised CLI/API response could turn this into a phishing flow.

Added a rule to allowlist the host of any approval URL (teams.splits.org / app.splits.org) before showing it to a human — anything else gets withheld with a warning instead.

> * The skill allows threshold-1 accounts where the agent auto-submits once it signs. Please make threshold ≥2 the default everywhere, and require an explicit user statement before creating/using threshold-1 accounts.

Made threshold ≥2 the default everywhere the agent is a signer, requiring an explicit user statement before creating/using a threshold-1 account. Threshold-1 accounts in the subaccounts table table are automation accounts (swap/sweep/buyback), which are unilateral by design (Bankr agent is not a signer on them). The table now separates "agent-signer" accounts (default 2) from "automation" accounts so the two don't get conflated.

PR #512, opened 2026-06-28
Adds a new provider `x402agentic` with one skill, `precheck`.

aipin-io:
**What it does:** a pre-pay risk gate for autonomous agents. An agent calls it
right before settling any x402 payment and receives an allow/warn/block verdict
with a 0–100 risk score and itemized reasons. Checks: envelope integrity, payee
address validity, denylist screen, community-flagged reports, price-vs-category
gouging, absolute ceiling, asset/network sanity, facilitator reputation, and the
caller's own spend policy.

**Hosting:** self-hosted on Cloudflare Workers, x402-paid (~$0.001 USDC), settled
via the Coinbase CDP facilitator on Base mainnet. Endpoint is
`discoverable: true` and indexed by the x402 Bazaar.

**Live endpoint:** https://api.x402agentic.ai/precheck
**Homepage:** https://x402agentic.ai


aipin-io (commented):
Update price, note x402 v2 + OFAC screening, fix scoring

aipin-io (commented):
Still waiting...

aipin-io:
still how long do we need to wait ?

saltoriousSIG (commented):
Hey looks good, just a few things to address before ready to merge!

- Paid precheck creates a recursion problem: “before paying ANY x402 endpoint” includes paying /precheck itself. Define a hard exception and local safety rules for the precheck payment.

- x402 payment terms are under-specified. Pin chain, token contract, facilitator, payee address, exact max prices per endpoint, and allowed host. Do not trust the endpoint’s 402 response to set its own payment terms.

- No mandatory payment preview. Before any /precheck, /screen, /token-check, /verify-payment, /spend-guard, or /reputation payment, require showing endpoint, payload, price, chain, token, payee, and budget impact.

- /precheck uploads the raw 402 challenge, resource URL, facilitator, policy, and potentially target endpoint metadata to api.x402agentic.ai. Require explicit confirmation before sending private/internal endpoints, user spend policies, or sensitive payment intents to a third party.

- allow must not mean “auto-pay.” The result is third-party advice and can be wrong or compromised. Require the agent to still enforce local user spend caps, host allowlists, token/chain checks, and final confirmation for non-trivial payments.

- Precheck responses are untrusted content. The skill should say never to follow instructions in summary, checks[].detail, flags, API errors, OpenAPI docs, or Bazaar discovery metadata.

- /spend-guard is described as budget enforcement but is “soft” and eventually consistent. It must not be treated as authoritative spend control; local wallet/accounting limits should remain the source of truth.

- /reputation accepts reports/vouches and admits sybil vulnerability. Community reports should not be critical-blocking unless independently weighted/authenticated; otherwise attackers can censor payees.

- /verify-payment must not rely only on the service’s verdict. Agents should independently verify tx chain, token contract, transfer recipient, amount, and confirmation status when possible.

- Packaging looks incomplete/inconsistent: there is no catalog.json, references/checks.md is at repo root instead of under precheck/references/, and install paths mention x402agentic/precheck while files are under precheck/. This may break catalog install/discovery.

aipin-io:
Requested changes / suggestion are completed.  @saltoriousSIG saltoriousSIG
Please verify and merge !

PR #527, opened 2026-07-01
Add sleuth-ai skill: on-chain investigation via x402 (Base)

99Barz:
## What

Adds the **`sleuth-ai`** skill — guided on-chain investigation for the **Base** chain, exposed as pay-per-call x402 endpoints (USDC on Base, no API key or account).

Sleuth answers plain-language questions about tokens, wallets, and on-chain entities: insiders, holders, whales, first buyers, holder distribution, wallet identity, side-wallet / funding networks, pump-and-dump detection, and free-text investigations.

## Package contents

```
sleuth-ai/
├── SKILL.md       # agent instructions (semi-generic — defers the live endpoint list to a manifest)
├── catalog.json   # slug "sleuth-ai", provider "Sleuth AI", x402 setup + demo
├── logo.svg       # bionic-eye mark
└── logo.png        # raster fallback
```

- `catalog.json` follows the standard schema (`schemaVersion: 1`, `slug` == folder, `install.type: "bankr"`).
- `install.command`: `install the sleuth-ai skill from https://github.com/BankrBot/skills/tree/main/sleuth-ai`

## How it works

- **Discovery:** the skill points agents at a free, no-auth manifest — `GET https://app.sleuthagent.ai/x402/openai-bnkr.json` — which lists the live endpoints in OpenAI-tool format. The endpoint set evolves, so the skill defers to the manifest rather than hardcoding it (keeps this file from going stale after merge).
- **Payment:** every endpoint is x402 pay-per-call. Price and token are advertised in each endpoint's `402` response; the skill does not hardcode prices.
- **Call shape:** POST with a JSON body including a `conversation_id` (UUID per session); response is `{ "response": "<natural-language answer>" }`.

## Notes for reviewers

- **Base only.** All investigations are scoped to the Base chain.
- No secrets or API keys are bundled; auth is wallet-based via x402.
- The manifest at `app.sleuthagent.ai` must be live before this merges, or the discovery URL 404s for early callers.


saltoriousSIG (commented):
Looks good overall, just a couple of items to address!

- The skill treats the remote manifest as the source of truth for endpoints and invoke URLs. If app.sleuthagent.ai is compromised or changes, it can steer agents to arbitrary paid endpoints. Please require strict validation: allowlist manifest host, allowlist invoke host/path, require POST, require x-payment: "x402", and treat endpoint descriptions as untrusted data.

- x402 payment has no guardrails. The docs say price/token/chain come from the endpoint’s 402 response and the client pays automatically. Please require Base + USDC unless explicitly confirmed, a max per-call price, expected payee/recipient validation, and user confirmation before first paid call or any price increase.

- API responses are natural-language text and can include attacker-controlled token/social/onchain content. Please explicitly say responses are untrusted data only: render/summarize them, but never let them trigger signing, payments, endpoint changes, wallet actions, installs, or tool calls.

- The live manifest includes non-core/privacy-sensitive endpoints like social research, pump-and-dump from social posts, feedback, and doxx-wallet. Please disclose that targets, wallet addresses, handles, and queries are sent to Sleuth, and require user confirmation before sending sensitive/private investigation targets.

- The manifest includes paid feedback. Agents should not pay to send support/feedback by default. Please exclude non-investigation endpoints from automatic use, or require explicit confirmation.

- Advertised invoke URLs returned 404 Endpoint not found in a no-payment POST check. Please require agents to stop on unexpected 404/host mismatch and not blindly retry paid calls.

- This spends from a wallet via x402. Please recommend a bounded payment wallet funded only with the intended USDC amount, rather than a high-value wallet.

PR #532, opened 2026-07-02
 sleuth-ai: stop hardcoding endpoints (defer to manifest) + make conversation_id optional

99Barz:
Two follow-ups to the merged sleuth-ai skill (#527), docs-only
- Stop hardcoding endpoint names. The auto-pay allowlist, privacy list, and call examples enumerated specific endpoints, which go stale as the set changes. Now fully discovery-driven: the pinned manifest is the only endpoint inventory; auto-pay is bounded by the payment pins + endpoint purpose, and the privacy gate keys on target content, not endpoint name.
- conversation_id is now optional. Omit it for a one-shot investigation, or reuse a stable id across calls to continue a session.

### DIFF
--- a/sleuth-ai/SKILL.md
+++ b/sleuth-ai/SKILL.md
@@ -79,8 +79,7 @@ natural-language investigations. Among other things it can:
 
 The manifest usually advertises a general free-text / ask-anything endpoint that answers any
 plain-language on-chain question — reach for whichever endpoint the manifest exposes rather than a
-fixed name. Each paid call is a single-shot investigation returning
-`{ "response": "<natural-language answer>" }`.
+fixed name. Each paid call returns `{ "response": "<natural-language answer>" }`.
 
 ## Payments (x402) — USDC or SLEUTH only, max $1 per call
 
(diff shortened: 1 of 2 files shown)


saltoriousSIG (commented):
Looks good, just  a couple of things to address before merging!

- Removing the explicit auto-pay endpoint allowlist increases risk. The new rule asks agents to judge endpoint “purpose” from manifest name + description, but those fields are provider-controlled remote text. Please keep a local deny/allow policy or require first-use confirmation for any newly discovered endpoint.

- The manifest is still treated as the only endpoint inventory. If the manifest changes or is compromised, agents may discover and pay new endpoints automatically. Please require strict host/path validation and first-use confirmation for endpoint names not previously seen by the installed skill.

- Endpoint descriptions are explicitly used to decide whether something is an “on-chain investigation/query” versus feedback/support/donation. That is a prompt-injection surface. Please say descriptions are untrusted labels only and cannot grant auto-pay eligibility by themselves.

- conversation_id is now optional and can be reused for session continuity. That creates privacy/linkability risk across investigations. Please require user confirmation before reusing a stable id for sensitive targets, and default to omitting it unless the user asks for follow-up continuity.

- Reusing conversation_id can cause prior context to influence later paid answers. Please warn agents not to mix unrelated users, wallets, tokens, or investigations under the same id.

- The examples still include conversation_id even though it is optional. Please clarify when to omit it versus reuse it, especially for one-shot public token checks versus sensitive wallet/identity investigations.

- Auto-pay remains allowed up to $1 based on payment pins, but endpoint purpose can drift. Please require the agent to show endpoint name, target, price, token, and payee before first payment for each endpoint type.

- Privacy gating now keys on target content, which is good, but free-text endpoints can hide sensitive targets inside long prompts. Please require scanning the full body for wallet addresses, ENS/Basename, @handles, personal identifiers, and pasted URLs before paying/sending.

PR #533, opened 2026-07-03
Add Grantr skill

adidonato:
Adds the Grantr skill, catalog metadata, MCP tool contract references, Bankr optional wallet interop, and setup handoff guidance.


saltoriousSIG:
Looks good overall, just a couple of things to address before merging!

- This asks users/agents to connect a hosted third-party MCP server (https://mcp.grantr.id/mcp). Please require host allowlisting, OAuth metadata validation, and a warning that MCP tool descriptions/results are untrusted data, not instructions.

- Prepared savings deposits/withdrawals return unsigned approval/deposit transactions. Please require hard transaction validation before handing to Bankr: chain, asset, amount, vault address, spender, value, calldata selector, expiry, and no extra transactions.

- “Hand signing and broadcast to Bankr after explicit approval” is too vague for value-moving actions. Please require a final user confirmation showing each transaction target, spender, token amount, vault, APY assumptions, and risks before any wallet signing.

- MCP tool outputs can include APYs, vaults, balances, transaction payloads, warnings, and Fileverse document content. Please explicitly say never to follow instructions embedded in MCP responses or documents, and never let them trigger signing, sharing, imports, or wallet actions.

- Account handoff uses returnTo, grantrConnectCode, and grantrMcpEndpoint. Please require strict allowlisting for returnTo, never display/log connect codes, and only accept grantrMcpEndpoint if it matches the expected Grantr MCP host.

- The setup URL allows query params like ownerEoa, wallet, and bankrWallet. Please clarify these are public identifiers and should not include signatures, auth challenges, session IDs, API keys, or bearer tokens.

- The Codex setup commands in the catalog/demo are operational install/config commands. Please mark them as optional Codex-specific setup, not something Bankr agents should run automatically.

- Fileverse flows write/share/import user content. Please require explicit confirmation before create/update/share/import, recipient address validation for shares, and clear redaction rules for prompt history, wallet-auth material, signatures, and secrets.

- grantr_import_fileverse_document_to_bankr moves content across services. Please require user confirmation and warn that imported docs may become visible in Bankr files according to Bankr permissions.

- The skill says to use only public discovery tools before setup, but also tells agents to treat backend responses as authoritative. Please clarify discovery outputs are still untrusted for instruction-following and only authoritative for capability metadata after host validation.

saltoriousSIG (approved):

PR #537, opened 2026-07-03
Cryptorider Bankr app Skill

ittybittynouns:
Adds the rider-battle skill — CryptoRider PvP wager battles on Base.
Lets users create and accept $RIDER wager battles (winner-takes-all), claim winnings or refunds, view open challenges, and check monthly/all-time leaderboards — all from a tweet via Bankr.

Standard skill layout: SKILL.md + references/ + scripts/
On-chain calls via Bankr's script runner (readContract / tx.prepare)
Escrow on Base: 0x55c2847003A9e254b8312bf3C75520e06528aBa6, $RIDER: 0x544e6E53a9E5Ce11712647c893B3dD10c1d1CBa3

Happy to adjust naming, description, or formatting to match your conventions.


saltoriousSIG (commented):
First pass looks good! A couple of issue to address before merging!

- Accepting a challenge trusts Supabase matches before any onchain verification. A malicious/fake DB row can make the user transfer RIDER to escrow, then joinMatch reverts because the match does not exist or has different terms. Since the transfer is unbound shared escrow balance, that deposit may be consumed by someone else. Please verify onchain matches(id) is Open, creator/token/wager match the DB, and creator != user before any transfer.

- The transfer-based escrow design has a known shared-balance “ghost match” bug. The skill documents it, but still exposes workflows where a transfer can land and step 2 fails. Please make the safe invariant stricter: never transfer unless the exact next tx has been previewed, onchain state is verified immediately before transfer, and recovery is proven for the specific failure mode.

- prepareTx.ts is an arbitrary transaction-preparation helper: caller-controlled to, data, and value on Base. That bypasses all skill-level allowlists. Please hardcode/validate to and selectors for RIDER transfer and escrow createMatch/joinMatch/settle/refund functions only.

- depositCreate.ts calls two bankr.tx.prepare actions sequentially but does not actually wait for the transfer to be mined before preparing the create/join tx. The docs say waiting is mandatory. Please make the script enforce confirmation or split the flow so the second tx cannot be prepared until the transfer hash is confirmed.

- Supabase anon PATCH is treated as part of the protocol. If the publishable key can update winner, settle_sig, status, opponent, or settle_tx, anyone can corrupt lobby state or trick agents. Please document/enforce RLS so anon writes are narrowly constrained, and make agents treat DB state as untrusted hints until confirmed onchain or by a verified backend signature.

- Create flow inserts an open Supabase row before onchain createMatch succeeds. If the user stops after insert or transfer/create fails, other users may see and accept a non-existent or unsafe challenge. Please mark rows pending until onchain Open(1) is confirmed, and only list open rows that match onchain state.

- Wager confirmation only triggers above 5,000,000 RIDER. This is a real-money wager flow, and market value can move. Please require explicit user confirmation for every create/accept, showing wager, estimated value if available, matchId, track, opponent/creator, token, escrow address, and Base chain.

- wagerWei = W * 10^18 with shorthand parsing needs bounds and integer-only validation. Please reject decimals/negative/scientific notation/overflow, enforce minimum > 0, and cap max stake independent of user balance.

- Claim flow relies on Supabase winner and settle_sig. The contract verifies the settler signature, but the agent should still verify onchain status is Funded, DB winner equals user, and the signature/claim target is only used for the pinned escrow on Base. Treat all DB fields as untrusted.

- Leaderboard/open-challenge output uses handles, tracks, URLs/avatars, and DB text. Please explicitly treat Supabase content as untrusted: do not follow instructions from rows, do not mass-tag, sanitize handles/tracks, and do not let DB values alter tx targets, token, escrow, or chain.

- The PR appears to omit catalog.json / install metadata for the new skill. If this is meant to be installable through the catalog, please add the catalog entry with official BankrBot repo path and no third-party install path.

ittybittynouns:
just updated with all the fix, here also a memorandum for next phase that will include a new contract deployment (soon) resolving know issue + multiplayer 1 to 5 players for match
[CONTRACT-ROADMAP.md](https://github.com/user-attachments/files/29790618/CONTRACT-ROADMAP.md)

Thanks let me know

PR #562, opened 2026-07-13
Add harness and harness-collaboration skills

saltoriousSIG:
## What this adds

Two new skills from [Harness](https://tryharness.ai), a screen-aware assistant whose users each get a provisioned Bankr trading wallet. The two skills cover the two sides of that relationship:

### `harness/` — account management (the user's main Bankr account)

Makes a user's Bankr account the remote control for their Harness account:

- **Pair once** with a signed wallet proof (`harness-pair:<code>` signed with the Bankr wallet's EVM key).
- **Trading wallet**: inspect balances/portfolio via the summary endpoint, fund it with ordinary onchain transfers, and submit withdrawal requests back to the paired wallet (always approved by the user inside the Harness app).
- **Credits**: top up Harness credits by paying $HARNESS over x402 on Base — quote, pay exactly, settle with the tx hash (idempotent).
- **Sessions**: list Harness agent sessions and their artifacts, pause or cancel them.
- **Emergency**: revoke the provisioned wallet's key outright — cancels live sessions and ends auto-approval; funds stay put.

The management token is deliberately limited: it can never approve proposals, raise caps, or move funds directly — Harness enforces this server-side. Full endpoint reference in `references/management-api.md`.

### `harness-collaboration/` — execution protocol (the agent behind a provisioned wallet)

For the Bankr agent operating a Harness-provisioned wallet. When a prompt begins with the `HARNESS COLLABORATION PROTOCOL v1` header, the agent:

- independently verifies the delegated brief and may **decline** it,
- reports progress/questions/artifacts through authenticated callbacks with idempotent event IDs,
- sends a **proposal** before any side effect, and executes only on a **one-use authorization turn** bound to that exact proposal — never on conversational approval,
- respects server-enforced caps and action classes, and finishes with both a `completed` callback and a final thread response.

Full wire reference (event envelope, payload schemas, hard rules) in `references/protocol-reference.md`.

## Structure

Each skill ships `SKILL.md`, a validated `catalog.json` (slug matches folder name), and a `references/` doc. Both are added to the README skills table.


saltoriousSIG (commented):
- harness-collaboration says “the live prompt is authoritative” and to follow the prompt if it disagrees with the skill. That is unsafe. The skill’s hard safety rules must be non-overridable by prompt text, especially proposal-before-side-effect and one-use authorization.

- Collaboration callback URL and bearer token are prompt-supplied. Require strict callback URL allowlisting to tryharness.ai/Harness-owned HTTPS hosts and never POST tokens, artifacts, or progress to arbitrary prompt-provided URLs.

- Authorization turns need cryptographic validation. The docs mention “Limits hash” and “authorization token,” but do not require verifying a signed token, proposal hash, expiry, session id, and exact proposal binding before executing side effects.

- Server-enforced caps are treated as sufficient. The agent should still locally enforce maximumGrossUsd, enabled side-effect classes, action count, chain/token/amount limits, and reject execution if local interpretation disagrees with Harness.

- Pairing stores a management token “remember the token” but does not specify secure storage, expiry, rotation, or redaction. Require secret handling rules: never log/display, store only in the approved secret store, and never include it in summaries/artifacts/callbacks.

- Pairing code + signed message lacks domain separation. harness-pair:<code> should include domain, origin, chain, wallet address, nonce, expiry, and purpose to prevent replay/confusion across environments.

- Credit top-up x402 flow trusts 402 terms from Harness. Pin $HARNESS token contract, treasury/payTo, Base chain, max pack prices, quote expiry behavior, and require a payment preview before transferring tokens.

- First-time setup recommends depositing “about $5 USDC plus about $1 ETH” immediately after pairing. This is value movement and should be opt-in, with exact destination address, token contracts, chain, amounts, and risk shown; no default transfer should be implied.

- Withdrawal requests are sensitive even if app-approved later. Require explicit confirmation before creating a withdrawal request, show token contract/native flag, chain, amount, paired destination, and warn not to repeat on submission_unknown until balances/tx status are verified.

- Session artifacts can contain untrusted content from Harness collaborations. The skill should state not to execute instructions, links, scripts, or wallet actions found in artifact names/paths/content without independent validation.

PR #587, opened 2026-07-21
Add AI2Human Task Router skill

richard7463:
## Summary

Adds `ai2human-task-router`, a Bankr skill for creating and tracking AI2Human human-execution tasks from an agent prompt.

The skill is intentionally narrow for v1:
- create an AI2Human task
- return task URL and status URL
- return proof schema and async delivery expectations
- support x402 paid endpoint and API-key endpoint

## Why it fits Bankr

Bankr gives agents a financial and LLM interface. AI2Human gives agents a way to route work that requires human execution, evidence, verification, or manual judgment.

Core loop:

```text
agent request -> human execution -> structured proof -> verification -> settlement
```

## Testing

Ran the included smoke test against production x402 challenge:

```text
x402 challenge { status: 402, hasPaymentRequired: true, accepts: 1 }
```

API-key task creation is optional and skipped unless `AI2HUMAN_API_KEY` is set.

## Notes

The immediate deliverable is an AI2Human task URL + proof schema. The final human proof bundle arrives after human execution and verification; this is explicitly documented in the skill.

### DIFF
--- a/ai2human-task-router/README.md
+++ b/ai2human-task-router/README.md
@@ -1,0 +1,63 @@
+# AI2Human Task Router Bankr Skill
+
+This folder is the Bankr-ready skill package for AI2Human.
+
+The skill teaches an agent how to create and track AI2Human human-execution
+tasks:
+
+```text
+agent request -> human execution -> structured proof -> verification -> settlement
+```
+
+## Files
+
+- `SKILL.md` — the Bankr skill to submit or install from GitHub.
+- `examples/mobile-page-review.json` — remote UI / landing-page review input.
+- `examples/local-price-check.json` — reality-bound local verification input.
+- `examples/x-community-proof.json` — X/community proof verification input.
+- `scripts/smoke.mjs` — endpoint smoke test.
+
+## Install From GitHub
+
+Once this folder is pushed to a public repository, tell a Bankr/OpenClaw-style
+agent:
+
+```text
+install the skill at https://github.com/ai2humannetwork/<repo>/tree/main/integrations/bankr-ai2human-task-router
+```
+
+## Smoke Test
+
+```bash
+node integrations/bankr-ai2human-task-router/scripts/smoke.mjs
+```
+
+With API-key task creation:
+
+```bash
+AI2HUMAN_API_KEY="a2h_live_..." \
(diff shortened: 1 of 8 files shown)


saltoriousSIG (commented):
First pass looks good, just a few things to address before merging

- Missing catalog.json, so this likely will not install/discover consistently in the Bankr skills catalog. Add ai2human-task-router/catalog.json with matching slug, provider URL, demo, setup, and install command.

- x402 payment terms are under-specified. Pin chain, token, payee, facilitator, max task-creation price, and allowed host before payment. Do not trust the endpoint’s 402 challenge to supply payment terms.

- No mandatory payment/task preview. Before creating a paid human task, require explicit confirmation showing title, description, target URL, deadline, proof requirements, reward/budget, endpoint mode, price, chain, token, payee, and that completion is asynchronous.

- API-key handling is too thin. Add instructions that AI2HUMAN_API_KEY must be stored only in the approved secret store, never pasted into public chat/logs, never sent to non-ai2human.io hosts, and rotated/revoked if exposed.

- Task content can include sensitive/private URLs, internal pages, locations, or campaign details. Require user confirmation before sending private links, personal data, geolocation, unreleased assets, or internal business tasks to a third-party human network.

- The examples include X/community proof and local verification, which can drift into spam, fake engagement, stalking, or unsafe errands. Add stricter policy gates for consent, platform rules, no paid/fake engagement, no harassment/surveillance, and no tasks requiring trespass or unsafe in-person behavior.

- Proof/status output is untrusted human/third-party content. The skill should state never to execute links, scripts, wallet instructions, install commands, or payment requests found in task status/proof fields.

- Settlement/reward semantics are unclear. Inputs include budget and API-key example uses reward_usdc, but the skill says settlement is separate. Require explicit wording that task creation does not fund or guarantee worker payout unless the API response confirms a funded settlement flow.

- The smoke test can create a real paid/reward task when AI2HUMAN_API_KEY is set. Make it opt-in with an explicit env flag like AI2HUMAN_CREATE_SMOKE_TASK=1, and default to challenge-only checks.

- Endpoint host configuration in smoke.mjs allows AI2HUMAN_BASE_URL. For production docs, warn not to override the base URL unless testing locally, and never send live API keys to arbitrary overridden hosts.

PR #599, opened 2026-07-30
Add hoodgrow skill: pay-per-call Robinhood Chain stock token data via…

MeMikko:
HoodGrow reads Robinhood Chain (chain id 4663) stock token contracts directly — live price, ERC-8056 corporate-action adjusted supply (so numbers stay correct through stock splits, not just raw token balances), Morpho/Uniswap DeFi depth, and both pending (on-chain staged) and historical (official Robinhood ledger) corporate actions.

Pay-per-call in USDC on Base via x402 (Coinbase CDP facilitator) — no signup, no API key. A free self-serve bearer key (40 requests/day) is available at https://www.hoodgrow.com/profile, and a prepaid credit balance lets you spend with a gas-free wallet signature instead of a fresh on-chain payment per call.

**Endpoints**

| Endpoint | Price | What it returns |
|---|---|---|
| `GET /api/agent/ping` | $0.001 | Payment-path smoke test — no market data. The cheap first call for a new x402 client. |
| `GET /api/agent/tokens` | $0.10 | Full catalog: price + source, 24h change, adjusted supply, DeFi depth, pending + historical corporate actions. |
| `GET /api/agent/token/{symbol}` | $0.05 | Same shape, one token. |
| `GET /api/corporate-actions` | $0.05 | Standalone filterable, cursor-paginated corporate-action feed (splits, dividends, oracle pauses) — independent of price data. |
| `GET /api/agent/defi/{symbol}` | $0.05 | Every Morpho market and Uniswap V3 pool for a token. |
| `GET /api/agent/holders/{symbol}` | $0.05 | Holder-count trend, 24h net supply change, top-holder concentration. |
| `GET /api/agent/slippage/{symbol}` | $0.05 | Price-impact estimate per pool for a USD-sized trade (read-only, executes nothing). |
| `GET /api/agent/ohlc/{symbol}` | $0.05 | OHLC candles with USD swap volume, up to 730 days. |
| `GET /api/agent/markets` | $0.05 | Top gainers/losers/volume/TVL across the catalog. |
| `GET /api/agent/trades` | $0.05 | Recent large ("whale") swaps with side, USD size, pool, tx hash. |
| `GET /api/agent/base/tokens` | $0.05 | Base-mainnet B20 native-equity registry with hourly DEX stats. |

**Access model** — x402 pay-per-call (30 req/min, no key), a free key (40/day), prepaid credits ($10→$11, $50→$60, $200→$260, spent via a signed header, webhooks billed $0.15/event), or a Builder key (300 req/min, no daily cap, webhooks included).

**Retries are safe if you ask for it** — send an `Idempotency-Key` header and a repeat of the same call replays the stored 2xx for 24h instead of paying again. Both SDKs take it as an opt-in argument (`idempotencyKey` / `idempotency_key`); it is not sent automatically. The credit-bundle purchase is deliberately not covered — never blind-retry that one.

**Clients** — `npm i hoodgrow`, `pip install hoodgrow`, or `npx hoodgrow-mcp` for an MCP server exposing every read endpoint as a tool. Docs: https://docs.hoodgrow.com

The skill file states the payment-safety invariants an agent must verify locally before signing (Base 8453 only, USDC only, fixed payee, per-endpoint price ceilings), tells the agent to treat API responses as untrusted third-party data, and documents the null contract on market stats — null means "not observed", never a fabricated zero.


MeMikko (commented):
:)

PR #602, opened 2026-08-01
bankr: sync skill with shipped platform changes

sidrisov:
Weekly sync of the `bankr` skill against shipped platform behaviour. Docs-only — no code, no schema changes.

## Token launching

**Fee structure.** The skill claimed a flat `0.7% swap fee split 95% creator / 5% protocol`. The real structure is a **0.7% pool swap fee with 95% going to the creator** (0.665% of trading volume, paid directly and claimable anytime), plus hook-added legs — the Bankr protocol fee, the BNKR buyback, and an LP fee — for **1.75% all-in**:

| Recipient | Share of volume |
|---|---|
| Creator — 95% of the 0.7% pool swap fee | 0.665% |
| LP fee (via hook) — compounds as locked liquidity in the token's own pool | 0.285% |
| Bankr protocol fee (via hook) | 0.475% |
| BNKR buyback (via hook) | 0.2375% |
| Protocol (Doppler) | ~0.0875% |

Because the LP fee compounds inside the token's own pool and never needs claiming, the creator side totals 0.95% of volume. The skill now also states that **fee schedules are fixed at launch** — earlier tokens keep the schedule they launched with, so integrations shouldn't retroactively re-derive them.

**Quote-only fees.** New launch option, previously undocumented in the skill: collect the entire creator share in the quote token instead of a mix of the launched token and quote token. Same total take — it's a denomination choice. Covered across all four entry points (natural language, `bankr launch --quote-only-fees`, `quoteOnlyFees: true`, web), plus the two knock-on effects for integrators: claims resolve against the hook contract's fees manager rather than the pool initializer, and a direct on-chain `updateBeneficiary` must target the hook address.

## Wallet API swaps

The swap surface grew several fields the skill didn't mention.

- **Request:** `slippageBps` (10–2000, default 500), `quoteId` for quote reuse, and `idempotencyKey` as a duplicate-submit guard.
- **Response:** documented the optional fields (`feeBps`, `swapImpactBps`, `maxPriceImpactBps`, `networkCostsUsd`, and the rest) with the important caveat that **only `from`, `to` and `minBuyAmount` are guaranteed** — everything else depends on the pricing venue and should be treated as `null` when absent, not hard-blocked on.

Two behaviours integrators reliably trip on, now called out explicitly:

- `slippageBps` always shapes the quote's `minBuyAmount`, but the **execution re-quote is clamped to 2%** on the aggregator path. A 2000 bps quote does not execute at 2000 bps there.
- **`quoteId` reuse only applies to same-chain EVM.** Cross-chain and Solana executions route before the stored-quote lookup and silently ignore it.

**Error and retry semantics.** Full status table (`400`/`403`/`409`/`429`/`502`/`503`/`504`) sorted into pre-broadcast, may-be-on-chain, and terminal buckets. The load-bearing warning: **never blind-retry a `504` or a LaunchLab `502`** — both mean the transaction was broadcast and only the confirmation is missing, so an automatic retry can execute a second swap. Also distinguishes the two price-impact rejections, since they read alike but mean opposite things: `400` is the venue refusing the trade, `403` is the caller's own wallet protection.

Finally, **tokenized-stock quotes are not location-gated** — only execution is. A successful quote is not clearance to execute, and the skill previously implied otherwise.

## Safety

- Price-impact protection gates on the fee-exclusive pool impact (`swapImpactBps`), not the display estimate, and the wallet's ceiling is readable from the quote as `maxPriceImpactBps`.
- It **fails closed** on venues that report no impact figure. Concretely: a brand-new Solana token still on its bonding curve can't be checked, so that fallback is refused rather than filled unguarded — those swaps are rejected by design while the limit is on.
- Spend limits apply on **every** swap path, including cross-chain and Solana legs.

## Other surfaces

- **BNKR staking is withdraw-only.** New deposits are deprecated; asking to stake returns an explanation rather than a transaction. The exit path (cooldown → redeem → withdraw) is intact. Flows that assume a deposit will go through need updating.
- **Hyperliquid funding is venue funding, not bridging.** Deposit and withdraw are separate operations with fixed destinations — a deposit only reaches Hyperliquid, a withdrawal only lands on Arbitrum. Moving funds further composes as a second cross-chain swap. The skill now advises naming the venue rather than saying "bridge", since a generic verb with a non-Hyperliquid destination is a request for that destination.
- **Polymarket** positions surface unspent collateral left in the deposit wallet. Worth documenting because `$0.00 claimable` is a truthful answer that can still hide money — the collateral was never staked, so it isn't redeemable, but it is recoverable by sweeping.

## Verification

- Fee table sums to 1.75% ✅
- No stale `95% creator / 5% protocol` or `1.2% swap fee` claims remain ✅
- Every documented parameter, field, status code and constant checked against current behaviour ✅

---
_Generated by [Claude Code](https://claude.ai/code/session_0153B6CqeM3i8H3LAsRnQeXa)_


Tjcinema:
Gm

PR #615, opened 2026-08-06
Add Teslr skill

HeresMyGit:
## Summary

- add a standalone Teslr skill for reading and safely controlling supported Tesla products
- include Bankr catalog metadata, setup guidance, and usage examples
- use the official Teslr PNG logo

## Validation

- `quick_validate.py teslr`
- validated `catalog.json` fields and repository path
- confirmed the contribution contains exactly three files


saltoriousSIG (changes requested):

- The advertised safe request pattern uses a single-quoted `node -e` program but tells the agent to replace its path and JSON body with API- and user-controlled values. Vehicle, energy, and selection IDs, place queries, shared URLs, plate text, and similar values can break shell quoting or alter the generated source. Please forbid interpolation of response or user data into command text, validate IDs, resources, and commands against strict local allowlists, and pass query/body values through a shell-safe encoded data channel that static Node code parses.
- State-changing native operations are explicitly sent with `confirm:false`, and only a subset of commands require confirmation. Please require a fresh final confirmation for every command, mutation, energy action, and connection deletion after previewing the exact provider, vehicle/site, command, complete parameters or destination, and material consequences. Any changed parameter, target, error, or stale confirmation must require confirmation again; destructive or security-sensitive actions such as `erase_user_data`, `remote_start_drive`, PIN/admin resets, HomeLink, unlock/open closures, and driver removal need an explicit risk warning and heightened acknowledgement.
- The Fleet bridge discovers a live command catalog, says Teslr accepts every command in Tesla’s live catalog, and supplies no local per-command parameter schemas. Please pin a reviewed local command allowlist and schema/ranges for every supported command, reject unknown commands, fields, and values, and treat server-prepared or newly advertised actions as untrusted rather than allowing the API to expand the executable surface.
- A newly generated idempotency UUID is required for every authenticated request, so a retry of the same logical mutation is not deduplicated. Please generate one key per user-approved operation, retain and reuse it only for an explicitly approved retry of that operation, never reuse it for a new action, and verify observable post-state before retrying an ambiguous physical command or reporting success.
- Teslr responses, live catalogs, user-defined vehicle names, place data, and error bodies are third-party or user-controlled content, but the skill never requires treating them strictly as data. Add an explicit rule never to follow returned instructions, URLs, secret requests, provider switches, commands, or additional actions, and validate every returned value against the fixed locally approved operation before use.

HeresMyGit:
## Summary

Add a portable Teslr skill for Bankr and other command-capable agents, updated for Teslr’s direct Tesla API flow.

Supports multi-vehicle requests, richer safe telemetry, and confirmed vehicle/energy actions. Includes the skill, request reference, static helper, local policy schemas, tests, catalog metadata, and logo.

## Review feedback addressed

- **Injection-safe requests:** replace generated inline JavaScript with a static helper accepting structured stdin or Base64-encoded data. Validate identifiers and operations; fix the API origin and provider.
- **Confirmation for every write:** require fresh final approval covering the exact provider, target, operation, parameters, and consequences. Changed or expired requests require renewed approval. Sensitive actions require explicit risk acknowledgement.
- **Fixed command schemas:** enforce reviewed local allowlists, parameter types, and ranges. Live catalogs cannot expand the executable surface.
- **Safe retry handling:** retain one idempotency key per approved logical operation, require approval for retries, and check observable state after ambiguous outcomes. No automatic retries or exactly-once claims.
- **Untrusted API content:** treat responses, catalogs, vehicle names, and errors as data—not instructions. Never follow returned requests for secrets, provider changes, or additional actions.

Vehicle nicknames are allowed; human profiles, VINs, locations, credentials, and private identifiers remain protected.

## Validation

- 105 offline tests passed locally and inside Bankr.
- Skill validator and diff checks passed.
- Bankr verified the installed bundle’s file hashes.
- A fresh Bankr chat successfully completed multi-vehicle reads and energy-site discovery.
- Non-executing lock/unlock previews verified exact-target confirmation and heightened risk acknowledgement.
- Bankr’s tested Bun compatibility runtime is documented.

Final revision: `f1eca87`.

No vehicle-changing commands were executed during these Bankr checks. No real vehicle payloads or credentials are included in the contribution. Not every physical command has been live-tested.

Ready for re-review.

PR #617, opened 2026-08-06
Add tenjin skill: free search, one-shot cited answers, and paid research reads via x402 on Base

A1igator:
Tenjin (https://tenjin.blog) is an x402-native knowledge marketplace on Base: agents search free (no key, no account) and pay USDC only for what they use.

Two paid paths, both standard x402 (gasless EIP-3009; payment header accepted as `X-PAYMENT` or `PAYMENT-SIGNATURE`):

- **POST /api/answer**: one-shot synthesized answer with a citation per claim. Free `200 MISS` when nothing fits, so an agent is never charged for a failure; each citation carries the payable URL of its source piece.
- **GET /api/read/...**: buy a full research piece (typically $0.01-$0.50) after inspecting its 402 preview and machine-readable answer card (questions answered, scope, exclusions, as-of date) for free. Wallets that already bought re-read free via the advertised `sign-in-with-x` extension; a re-pay attempt from an entitled wallet is refused with `409 already_purchased`, never double-charged.

**Deterministic guardrails.** Three bundled read-only scripts enforce the review's requirements in code rather than prose; none touches keys, signs, or pays:

- `scripts/verify-402.mjs` (dependency-free): mandatory pre-payment gate. Enforces the registry pins plus payee validation - `/api/answer` must pay the pinned Tenjin treasury exactly; reads must pay a contract whose on-chain bytecode byte-matches the 0xSplits PushSplit proxy every Tenjin split shares (EOAs and unknown code rejected). Rejects downgrade attempts (more than one `accepts` entry, or any `extra` field beyond the USDC EIP-712 domain). Keeps a local attempt ledger so a purchase can never be signed twice, and verifies settlement against the mined USDC transfer (auto-discovering the tx for wallets like the Bankr CLI that do not print it).
- `scripts/tenjin-api.mjs` (dependency-free): every free, keyless call as a command; a 402 result prints the exact verify command for its quote.
- `scripts/siwx.mjs`: CAIP-122 signing hygiene via the official `@x402/extensions` package (pinned, lockfile committed). Validates a 402's advertised signing fields against the pins before any message exists, takes the signature via stdin only, sends it only to the pinned origin, and single-uses each context.

**Consent rides Bankr's own surfaces**: `bankr x402 call` confirms interactively with `--max-payment` pinned to the verified price (`-y`/`--ni` reserved for the user), and publishing defaults to a final-payload confirmation, with autonomy only via durable user-authored agent configuration.

`x402-registry.json` ships the strict signing policy the scripts implement: pinned origin, Base mainnet, canonical USDC, $1 amount cap, the pinned treasury for answers, and a payee policy for reads (required bytecode match, plus a documented maximum check that the split's recipients are the piece creator and the pinned treasury at the 10% platform fee).

Tested live on Bankr rails with this revision, through the scripts exactly as written: free search and inspect, verify PASS with tamper and downgrade rejections (wrong origin, tampered payTo, EOA payee, permit2 extras, extra accepts entries), a real $0.05 read settled by `bankr x402 call` (tx `0x0be2ca10f10bfa908557ddc3bbd4265ccc8f1f874a09520387057dcc79514b70`) and verified on-chain by the reconcile step, the double-payment guard refusing a re-buy, a free SIWX re-read of the owned piece signed via `bankr wallet sign` (HTTP 200, no charge), a draft publish (201) and delete (204) from the same Bankr wallet, and a cold fresh-agent run of the free flows whose documentation findings are folded into SKILL.md.

Live contract: https://tenjin.blog/llms-full.txt · OpenAPI: https://tenjin.blog/openapi.json


saltoriousSIG (changes requested):

- Paid reads and answers can enter an x402 auto-pay retry without final user confirmation. Please require a payment preview and explicit confirmation before each paid call unless the user has configured a local autopay policy. Show the exact endpoint, Base chain ID, USDC contract, atomic and human amount, payee, facilitator, authorization expiry, and local per-call/session spend remaining; API `maxPrice` and the registry’s global $1 cap are not substitutes for user consent or local spend limits.
- `payTo` is intentionally accepted from each 402 challenge and no facilitator is pinned. A compromised Tenjin origin could therefore redirect every allowed payment to an arbitrary recipient while satisfying the current checks. Please pin the facilitator and constrain payees to an independently verifiable allowlist or locally validated split-contract factory/implementation and derivation; reject EOAs, unknown bytecode, alternate settlement routes, or any recipient that cannot be validated.
- Retry and settlement handling is under-specified for paid requests, especially the paid `POST /api/answer`. Please require a unique local payment-attempt record, never blindly retry a paid POST or generate a second authorization after timeout, reject reused/stale nonces and validity windows, and verify the signed EIP-3009 fields match the confirmed terms. Before reporting purchase success, validate the `PAYMENT-RESPONSE`, mined Base receipt, successful USDC transfer, exact amount/payee, and returned entitlement; treat 409, timeouts, and ambiguous responses as reconciliation cases rather than new payment attempts.
- The skill directs agents to fetch `https://tenjin.blog/skills.md` for a CAIP-122 message recipe and then issue a `personal_sign`, but does not require local validation of the message. Please treat the remote recipe as untrusted data and require preview plus explicit confirmation of the exact message, pinned HTTPS domain/URI, wallet, Base chain, action and resources, fresh server nonce, issued-at/expiry, and narrowly scoped purpose before signing. Treat SIWX signatures as bearer credentials: do not log or persist them, never send them off-origin, and do not reuse them across reads or publishing actions.
- Publishing sends agent-produced research and optional `searchId` attribution to a third party without a mandatory disclosure gate. Please require a final payload preview and user confirmation before upload, with removal of private URLs, personal data, internal names, credentials, unreleased material, and content the user lacks rights to publish; remote trending/search results must not trigger autonomous publishing.

A1igator:
Thanks for the thorough review. All five items are addressed in the latest push (20b9584), and the theme of the revision is your underlying point: the guarantees are now enforced by deterministic, read-only scripts rather than asked of the model. Three scripts ship with the skill; none touches keys, signs, or pays.

**1. Payment preview and confirmation.** `scripts/verify-402.mjs` is now a mandatory gate before every paid call. On PASS it prints the full preview: endpoint, Base chain id, the USDC contract, atomic and human amount, the validated payee and its class, authorization expiry, and USDC already attempted today from its local ledger. Consent then rides the wallet's own flow: the skill instructs `bankr x402 call <url> --max-payment <verified price>`, which confirms interactively by default, and forbids `-y`/`--ni` unless the user themselves asked for autopay. The registry caps are documented as damage bounds, not consent.

**2. payTo and facilitator.** payTo is no longer open-ended. It has exactly two validated classes, declared in `x402-registry.json` → `payeePolicy` and enforced by the script: `/api/answer` must equal the pinned Tenjin treasury (`0x578c640E2D53023066026381B8CC362539741063`), and `/api/read/...` must carry contract code on Base byte-identical to the 0xSplits v2 PushSplit proxy every Tenjin split shares (pinned as `runtimeBytecode`; factory `0x8E8eB0cC6AE34A38B67D5Cf91ACa38f60bc3Ecf4`). EOAs, undeployed addresses, and unknown bytecode are rejected; Tenjin deploys each split before quoting it to buyers. The registry also documents the maximum check: the split's recipients are the piece creator's wallet plus that same pinned treasury at the 10% platform fee, verifiable at app.splits.org or via the on-chain splitHash.

On the facilitator: unlike the permit2/erc7710 skills in this catalog, where the signature authorizes a spender contract and pinning the facilitator is load-bearing, Tenjin's exact/eip3009 flow has no facilitator or spender in the signed object; the authorization's `to` is the payee itself, so payee and amount validation is what bounds a compromised origin. To make that non-downgradeable rather than assumed, the verifier rejects any challenge carrying more than one `accepts` entry or any `extra` field beyond the USDC EIP-712 domain `{name, version}`, so a tampered origin cannot steer a multi-scheme wallet into a spender-mediated flow these checks never see. Both rejections are exercised in the test evidence below.

**3. Retry and settlement.** The script maintains a local attempt ledger. Every PASS records a pending attempt first; a second verify of the same purchase fails until the first is closed with `verify-402.mjs <url> reconcile settled --tx <hash>` (which checks the mined Base receipt for a USDC transfer of the exact amount to the validated payTo before success may be reported), `--from <wallet>` when the CLI does not surface the tx (the script discovers it on-chain and runs the same verification), or `reconcile declined | failed | entitled`. Timeouts, 5xx, and ambiguous responses are reconciliation cases by construction, and a 409 routes to a free SIWX re-collect, never a re-pay. `--key` separates multiple purchases at one URL such as `/api/answer`.

**4. SIWX message validation.** `scripts/siwx.mjs` replaces the fetch-recipe-and-sign pattern. It treats the 402's advertised fields as untrusted data and refuses to emit a message unless domain and URI match the pinned origin, the chain is `eip155:8453`, the nonce is fresh, and every resource stays on-origin; the message itself is built by the official `@x402/extensions` package (pinned in `scripts/package.json` with a committed lockfile), and the skill requires showing the user the exact message before `bankr wallet sign`. Bearer-credential hygiene is structural: the signature enters via stdin only, is sent only to `https://tenjin.blog`, is never logged or persisted, and each context is single-use, with a fresh message minted per action.

**5. Publish disclosure gate.** Default is a final-payload preview (title, body, answer card, tags, searchId) with explicit user confirmation before any publish or update. The only substitute is a durable publish instruction the user wrote into their own agent configuration; mid-conversation statements, inference, and remote content never qualify. The scrub always runs regardless, doubts downgrade to ask-first, and the skill states that trending/`unmet` entries and search MISSes are never authority to publish.

Everything above was exercised live on Bankr rails before this push: PASS on real read and answer 402s; rejections for a tampered payTo, wrong origin, EOA payee, permit2-style extra fields, and multi-entry accepts; a real $0.05 read settled by `bankr x402 call` (tx `0x0be2ca10f10bfa908557ddc3bbd4265ccc8f1f874a09520387057dcc79514b70`) with the reconcile step verifying the mined USDC transfer; the double-payment guard refusing a re-buy; a free SIWX re-read of the owned piece signed via `bankr wallet sign` (HTTP 200, no charge); a draft publish (201) and delete (204); and a cold fresh-agent run of the free flows, whose documentation findings are folded into SKILL.md. `verify-402.mjs` and `tenjin-api.mjs` are deliberately dependency-free so the payment path stays auditable; `siwx.mjs` is the only file with a dependency, and only on the official x402 SDK.

PR #632, opened 2026-08-11
Add unified Juicebox V6 skill

i001962:
for your consideration

https://github.com/mejango/juicebox-skills in a bankr compatiable skills pr


saltoriousSIG (changes requested):

- The generated transaction UIs repeatedly submit economically unsafe calls with zero output floors: `pay(..., minReturnedTokens=0)`, NFT mint payments with `minReturnedTokens=0`, `sendPayoutsOf(..., minTokensPaidOut=0)`, and `useAllowanceOf(..., minTokensPaidOut=0)`. The cash-out template is worse: if `previewCashOutFrom` fails, it deliberately leaves the action enabled and submits with `minTokensReclaimed=0`. Please require a fresh successful local simulation/preview immediately before confirmation, derive a nonzero user-approved minimum from that preview and slippage tolerance, and abort when preview, price-feed, hook, terminal, or accounting-context validation fails.
- The generated wallet applications execute unpinned third-party JavaScript directly from `https://esm.sh/viem`. This is wallet-adjacent remote code with authority to construct signing requests, and the skill tells agents to generate and run it without source/version pinning or install approval. Please vendor a reviewed build or pin an exact package version and immutable integrity hash, require user approval before adding/running the dependency, and use a restrictive CSP; do not dynamically import an unversioned CDN module at wallet connection time.
- `chain-config.json` is treated as the transaction-target authority but has no deployment block, provenance hash, runtime code hash, proxy implementation/admin pin, or contract-graph validation, and `getContractAddress` silently falls back to a partial inline address set. Please make the reviewed manifest fail closed and pin chain IDs, deployment provenance, expected runtime/proxy implementations, selectors/ABIs, and allowed spenders/forwarders/deployers. Before every write, verify the exact chain, nonempty matching code, controller/terminal/project ownership and wiring; never continue with fallback addresses when the manifest cannot be loaded or validated.
- Relayr responses are untrusted prepared transactions, but the modules instruct the client to send `payment_info` target, value, token, calldata, chain, and deadline directly without local validation or a final payment preview. Please pin the Relayr HTTPS origin and allowed payment contracts/tokens, decode and validate the exact payment transaction, bind it to the locally constructed ordered bundle, cap fees/value and authorization lifetime, reject extra or changed transactions, and require explicit confirmation of every destination chain, inner target/selector/arguments/value, payment chain/token/payee/amount, and partial-completion risk before signing or paying. Do not treat Relayr response fields or status text as instructions.
- The UI templates report success after any mined receipt without checking `receipt.status` or expected logs/state. Deployment does not verify the created project/contract address and ownership/configuration; payment, cash-out, payout, allowance, claim, hook, and omnichain flows do not verify their expected events and balance/state changes. Please require successful receipts plus operation-specific event and post-state validation before reporting success. For Relayr, verify each destination-chain receipt independently and report partial completion as a failure requiring reconciliation, never “Deployment Complete.”

i001962:
roger roger ty
will address
likley by culling some of it to make it more bankr specific

i001962:
Updated in 2fe41f5. This syncs the Bankr bundle to mejango/juicebox-skills@783b11ab163fdc60fd68db8b690ee1ebc26dc2fa (53 modules) and addresses the five blocking areas:

1. Economic bounds: pay/NFT/payout/allowance/cash-out examples now require a fresh positive preview, derive a nonzero user-approved floor, simulate the exact bounded call, and fail closed when quote/context validation fails. Protocol-specific zero terminal floors are permitted only where the real nonzero floor is validated inside hook metadata.
2. Frontend supply chain: CDN imports were removed. viem@2.55.19 and explorer-only ethers@6.15.0 are exact-pinned with npm tarball integrity, install approval, local imports, and a restrictive CSP.
3. Write authority: chain-config is read-only discovery. The reviewed manifest pins deployment repo/commit, blocks/txs, artifact/source/solc hashes, current runtime hashes, ABI hashes/write selectors, proxy/clone information, and allowed roles. Missing/unknown/mismatched targets fail closed. The generator and verifier cover 276 canonical write targets on 8 chains.
4. Relayr: the manifest pins origin, prepayment(bytes16,uint40) contract/selector/code hash, native token, payment chains, value, and lifetime caps. The template rebinds the service record to the exact ordered local bundle, decodes the bundle UUID/deadline, verifies live code, simulates the exact payment, and requires a full destination/payment/partial-risk confirmation.
5. Completion proof: the shared receipt helper now requires receipt success plus an operation-specific event/post-state proof. Deployment examples verify created ownership/wiring or remain uncertain; Relayr verifies each destination transaction/receipt independently and treats missing proof or partial completion as reconciliation-required, never complete.

Validation run: manifest generation with live bytecode pins; independent `verify-reviewed-manifest.mjs --live` (including Relayr on all payment chains); JS syntax checks; catalog/frontmatter/router checks (53 routed modules); `git diff --check`; and scans confirming no esm.sh/jsdelivr/unpkg/script-src imports remain.

saltoriousSIG:
LGTM

PR #657, opened 2026-08-22
bankr: weekly sync — credit transfers, degen mode, /runs, Gemini 3.7 Flash default

sidrisov:
Weekly sync of the `bankr` skill against what actually shipped, plus a few pre-existing inaccuracies the pass turned up.

## Shipped this week

**LLM credits are transferable peer-to-peer.** New section in `references/llm-gateway.md` plus a summary in `SKILL.md`: `POST /llm/credits/transfer`, the agent phrasings that reach it, and the rules that bound it — recipient must be an existing Bankr user (X username or `0x` address, no ENS on this path), only *purchased* credit moves (grants aren't transferable), $1 minimum, $500 gross per trailing 24h, read-only keys refused, caller-supplied `transferId` for idempotency, and wallet-level pause / permitted-recipients controls applying as they do to on-chain sends. Documents the explicit failure codes and `transferableUsd` on `GET /llm/usage` so a client can offer a "Max" the endpoint won't reject.

**Gemini 3.7 Flash is the agent's default model.** Added to the model table and marked as the default; 3.6 Flash demoted to a previous-generation row. The Max Mode section named the old default and now names the new one.

**Degen mode for token launches.** New section in `references/token-deployment.md`, a bullet in `SKILL.md`, a parameter-table row, and a prompt example: a fixed $2,500 starting market cap, explicit opt-in by name only (a token *called* DEGEN doesn't opt you in), no custom cap, unavailable on partner deploys, and no CLI flag yet.

**Polygon pUSD → USDC.** Asking to convert pUSD to plain "USDC" on Polygon now resolves to `USDC.e` and takes the 1:1 Offramp unwrap instead of failing to route; the swap-semantics bullet says so, and notes the buy direction is untouched.

**Wallet-mode swap routing.** Swaps signed from an external/connected wallet now walk the same venue order as custodied ones, so a thin-liquidity pair reaches its direct-pool venue rather than being diverted to the bridge aggregator — which can collapse an approve-plus-router pair into a single transaction to sign. Also: a leg whose build fails is now reported as failed rather than handed over to sign under a success message.

**Impostor-token screening.** New safety subsection plus a `references/safety.md` rule. Airdropped dust reporting a canonical ticker (`USDG`, `USDC`, `USDT`, `EURC`, a native or wrapped symbol) from a non-canonical address now keeps its negative verdict and drops out of the portfolio listing instead of reaching the agent as a clean-looking entry. Notes the narrowness — genuine canonical tokens and anything bought through Bankr are never hidden — and gives the durable habit: name the ticker, don't trade an address pasted from a balance list.

**Run outputs (`/runs`).** New section in `references/files.md` and a bullet in the file-storage capability list. Covers the `./output/` auto-persist channel, oversized tool results being stored rather than pasted inline, and the limits: conversation-scoped, ~14 days refreshed on access, 10 MB per file, 100 MB per conversation, 25 files / 10 MB per sandbox call, text only.

## Gaps caught while syncing

These were already wrong before this week's changes; fixed here since the same sections were being touched.

- **The EVM launch default differs by surface.** The skill claimed Base everywhere. The CLI and the web launch form preselect Base, but the agent and the deploy API fall back to Robinhood Chain when no chain is named. Corrected in three places, with an explicit "name the chain" instruction.
- **Wrong deploy endpoint.** The stock-paired launch example used `POST /token/deploy`; the endpoint is `POST /token-launches/deploy`.
- **The LLM daily spend budget was undocumented.** Added: the cap is measured over a trailing 24 hours rather than a calendar day (nothing resets at midnight — capacity returns as charges age out), over-budget spending requests return `402` with `type: daily_budget_exceeded` (distinct from `insufficient_credits`), read-only `GET` endpoints stay unblocked so you can poll for recovery, and the same budget ends Max Mode agent runs early. The rolling-window measurement and the blocked-caller semantics landed this week; the underlying cap predates the window.

## Verification

Every limit, endpoint path, error code, and default in this diff was read off current source rather than inferred — model default, transfer minimum/cap/permission gates, degen-mode market cap and partner-deploy rejection, `/runs` byte and count caps and TTL, and the per-surface launch-chain defaults.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CTvGjcc8zq6Rgn71RDFdL2)_

### DIFF
--- a/bankr/references/llm-gateway.md
+++ b/bankr/references/llm-gateway.md
@@ -50,7 +50,8 @@ bankr config get llmKey
 | `claude-sonnet-4.6` | Anthropic | Previous generation Sonnet (1M context) |
 | `claude-sonnet-4.5` | Anthropic | Earlier Sonnet (1M context) |
 | `claude-haiku-4.5` | Anthropic | Fast, cost-effective (200K context) |
-| `gemini-3.6-flash` | Google | Latest Flash — the Bankr agent's default model (1M, image input) |
+| `gemini-3.7-flash` | Google | Latest Flash — the Bankr agent's default model (1M, image input) |
+| `gemini-3.6-flash` | Google | Previous Flash, coding and agents (1M, image input) |
 | `gemini-3.5-flash` | Google | Fast general-purpose (1M) |
 | `gemini-3.5-flash-lite` | Google | Ultra-fast, lowest cost (1M) |
 | `gemini-3.1-pro` | Google | Long context, reasoning (1M) |
@@ -177,7 +178,7 @@ Only a **trailing** tier token counts as the opt-in, so unrelated model IDs cont
 
 ### Max Mode — Choose the Agent's Model
 
-Max Mode replaces the Bankr agent's default model (`gemini-3.6-flash`) with any gateway model, billed per token from your **LLM credit balance**. It's the pay-per-use alternative to a Bankr Club subscription for unlimited terminal messages, and unlike Club checkout it works with external/connected wallets.
+Max Mode replaces the Bankr agent's default model (`gemini-3.7-flash`) with any gateway model, billed per token from your **LLM credit balance**. It's the pay-per-use alternative to a Bankr Club subscription for unlimited terminal messages, and unlike Club checkout it works with external/connected wallets.
 
 ```bash
 bankr agent "analyze my portfolio" --model claude-opus-5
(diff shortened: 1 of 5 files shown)


bankr/references/llm-gateway.md
```diff
@@ -244,6 +245,54 @@ spend order = expiring grants first (soonest-expiring), then the permanent pool
...
+
+Purchased credit is transferable peer-to-peer. Ask the agent, or call the API directly:
+
+```bash
+bankr agent prompt "Send $20 of LLM credits to @alice"
+bankr agent prompt "Transfer 5 dollars of my LLM credits to 0xRecipient"
+```
+
+```bash
+curl -X POST "https://api.bankr.bot/llm/credits/transfer" \
```
sidrisov, line 258: is this api dual auth which supports api key as well?

sidrisov: Yes — dual auth, same as the other credit endpoints. Checked the middleware rather than going from the example: it takes a web session token first if one is present, otherwise falls through to API-key auth, which reads either `X-API-Key` or `Authorization: Bearer`. The key needs LLM Gateway access enabled, and since a transfer is a write, read-only keys are refused on either path.

The example was accurate but left `X-API-Key` looking like the only option, so I've spelled it out in the line under the snippet (`60e92ed`).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #659, opened 2026-08-22
Update AZZLE skill for dual-market V2

azzle-lab:
## Summary
- Update the published AZZLE Bankr skill from the single-market V2 pin to reviewed standard and micro pins.
- Require market-scoped task IDs (`v2:standard:N` / `v2:micro:N`) and keep addresses in the bundled pins only.
- Remove the retired subgraph discovery helper and route read-only discovery through the first-party helper and API.

## Test plan
- [x] `npm run check:bankr-skill` passed in the AZZLE source repo
- [ ] Confirm `azzle/` still contains only the intended skill files plus `logo.svg`
- [ ] Confirm both pins declare `version` 2.0.0 and `chainId` 8453
- [ ] Confirm `v2-tasks.sh` rejects bare numeric IDs and requires an explicit micro selection
- [ ] Confirm the retired subgraph helper and single-market pin are gone

This PR updates the existing `azzle/` skill only. It does not modify unrelated skills.

Made with [Cursor](https://cursor.com)


saltoriousSIG (changes requested):

- The new micro signing graph and updated standard pin authenticate only addresses, version/chain, and nonempty self-consistent contracts. The bundle contains no approved ABI/selector allowlist or expected runtime/implementation code hashes, and the rules do not validate `deploymentBlock`, `deployer`, `bundleHash`, or `finalizedTx`; a wrong implementation can expose the expected wiring accessors and still pass these gates. Please pin the reviewed code identities and allowed selectors/ABIs for every signing target and spender, verify them from Base before each approval/write, and require the finalization receipt to be successful and match the exact factory, deployer, block, and emitted contract graph.
- `v2-tasks.sh task` and `scope` validate only the requested ID string, then trust and print the first-party response without checking its `id`, `market`, `chainId`, `registryAddress`, `escrowAddress`, or scope against the selected pin. A compromised or misconfigured API can show terms from one market while the later claim targets another. Please strictly validate the response schema and all market/address bindings, then re-read the task record and public scope directly from the selected pinned Base contracts before presenting or confirming a claim; fail closed on any mismatch.
- The production SDK path is not safely pinned or market-bound. It imports `@azzle/agents` without naming an exact reviewed version/integrity and calls `loadBaseMainnetV2Manifest()` without an explicit market, even though this PR changes the standard pin and adds micro. Please require user approval for installation, pin the reviewed package source/version and integrity, select the market explicitly from the validated task namespace rather than an ambient default/environment value, and compare the loaded manifest with the corresponding installed reviewed pin before constructing a wallet-connected client.

azzle-lab:
Thanks — addressed in 8627da4.

Code identity / allowlist. Each market now has a reviewed identity pin (runtimeCodeHash, USDC implementationCodeHash, deploymentBlock, deployer, factory, bundleHash, finalizedTx, and the SuiteDeployed hash). ./scripts/v2-tasks.sh verify <market> re-checks the successful finalization receipt and live Base code hashes before writes. ./scripts/v2-tasks.sh allow <target> <calldata> rejects any selector not in references/signing-allowlist.json.

task / scope. Those helpers parse v2:standard:N / v2:micro:N only, re-read tasks() and scopeOf() from the selected pinned contracts, then fail closed if an API payload disagrees on id, market, chainId, registryAddress, escrowAddress, or scope.

SDK pin. references/sdk-pin.json pins @azzle/agents@0.5.0 with integrity. Install requires explicit user approval. Market comes from parseTaskRef; loadMarketManifest(market) must match the installed reviewed pin before constructing a wallet client. No no-market loadBaseMainnetV2Manifest path.

PR #665, opened 2026-08-25
Add pantheon-staking — stake Pantheon vaults on Base by conversation

KoalaDragonKingKong:
Adds the Pantheon staking skill to the catalog.

What it does: lets any Bankr user stake into live Pantheon vaults on Base, view positions, and claim rewards by conversation. Token resolution is registry-only (never DEX search or user-supplied addresses), every stake requires a relayed lock/fee disclosure and explicit user affirmation before any approval, approvals are exact-amount only, and the skill contains zero transfer instructions — all actions are contract calls from the user's own wallet against our audited vault. Unstake/exit verbs are deliberately excluded in v1 and routed to the web app.

Tested: installed into a fresh Bankr agent from the canonical repo in one message; two real stakes executed end-to-end through the skill and verified on-chain to the raw unit; adversarial prompts (fake token address, skip-consent, unlimited approve, unstake) refused per the skill's rules.

Install type is external, pointing at the canonical repo we maintain and publish on pantheonvaults.com: https://github.com/PantheonVaults/pantheon-skill — this folder carries the same SKILL.md and references for the catalog listing.

Happy to adjust to the bankr install form if you prefer installs vendored from this repo.

### DIFF
--- a/pantheon-staking/SKILL.md
+++ b/pantheon-staking/SKILL.md
@@ -1,8 +1,8 @@
 ---
 name: pantheon-staking
-description: Stake Base creator coins into Pantheon vaults, view staking positions, and claim monthly rewards. Use when the user mentions Pantheon, pantheonvaults, staking a creator coin for yield, or checking/claiming Pantheon vault rewards.
-tags: [staking, defi, base, pantheon, creator-coins, yield]
-version: 1
+description: Stake creator coins into Pantheon vaults on Base and Robinhood Chain, view staking positions, and claim monthly rewards. Use when the user mentions Pantheon, pantheonvaults, staking a creator coin for yield, or checking/claiming Pantheon vault rewards.
+tags: [staking, defi, base, robinhood-chain, pantheon, creator-coins, yield]
+version: 2
 visibility: public
 metadata:
   clawdbot:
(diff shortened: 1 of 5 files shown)


saltoriousSIG (changes requested):

- The hosted registry is allowed to choose the transaction chain, vault/spender, token contract, and decimals: the skill explicitly says to take `chain_id` and `vault_address` from its response. A compromised registry could redirect an approval and stake. Please pin the only allowed mappings locally (`8453` → `0xBf52…8fDb`, `4663` → `0x541E…3edB`), reject redirects or any mismatch, strictly validate the response as untrusted data, and verify each selected token’s code, address, decimals, symbol, and active pool from the same chain before showing the exact contract and raw amount for confirmation. Registry text, symbols, URLs, or newly added entries must not authorize a wallet action.
- Stake and claim writes lack the required Bankr transaction safety gates. Before signing, require explicit wallet support for the selected chain and locally decode the actual prepared transactions: stake must be exactly an exact-amount `approve` followed by `stake(token,6,amount,false)`, with pinned chain/targets/selectors, zero value, exact token/spender/human and raw amounts, order, gas/fees, and no extra calls; claims must preview and confirm the exact reward-token addresses, gross/fee/net amounts, gas, and transaction count. Obtain final confirmation for the specific writes, stop on scanner errors, then wait for mined successful receipts and verify allowance consumption, stake/claim events, position/accounting, and received balances before reporting success.
- The contracts are upgradeable proxies, but the write flow never checks the EIP-1967 implementation or code hash and the mandatory disclosure hides the material admin risk. Base moved to V8 one day before this PR, while the Robinhood proxy and ProxyAdmin are controlled by one EOA; the current wording instead says the lineage is audited and that Pantheon never takes custody, even though tokens are transferred into an admin-upgradeable vault. Please pin and verify the reviewed implementation/code hash before every write, fail closed after any upgrade pending review, accurately state the audit status of each live implementation, and disclose that principal is locked in an upgradeable contract, admin compromise or rule changes can cause loss, and funds can be lost.
- Catalog installation executes a mutable external repository: both `command` and `reference` point to the default branch of `PantheonVaults/pantheon-skill`, and `SKILL.md` tells installed agents to reinstall from it. That lets unreviewed future commits replace these wallet instructions after catalog approval. Please vendor/install the reviewed `pantheon-staking` directory from this repository, or pin the external source to an immutable commit plus integrity digest and require explicit user approval before installation or update.
- The required registry is not currently usable as a machine-to-machine endpoint from the reviewer environment: both documented `GET ...?chain=base` and `GET ...?chain=robinhood` requests return HTTP 403 with a Cloudflare “Attention Required” page to `curl`. Because the skill has no safe fallback, discovery and every write are unavailable. Please expose an authenticated or public API route that does not require browser cookies/JavaScript or challenge non-browser agents, and add an install-time/live health check that fails closed on non-JSON or non-2xx responses.

PR #667, opened 2026-08-25
Quotient skill 1.9.0 — live signal board, first-use x402 budget, current API surface

jordan-c-olm:
## Summary

Updates the maintained `quotient/` skill (merged in #609) to 1.9.0, regenerated from Quotient's canonical skill source. The API contract moved from OpenAPI 5.0.0 to 13.9.0 since the merge; this refresh brings the skill current and fixes a live behavioral bug.

Note on provenance: PR #609's page is no longer publicly visible (the submitting account was flagged by GitHub, unrelated to this repo), so this PR ships from a fresh fork and restates the review follow-through below.

## The bug this fixes

A user asked the Bankr bot which Quotient signals were live and got **one** signal when ~14 were live. The merged skill routes "best signal" to `GET /signals/featured` (a single row by construction), its demo leads with `featured`, and its docs claim `/signals` defaults `window=24` — the API applies **no** window filter unless one is sent. 1.9.0 routes "what's live / which are best" to a bare `/signals` call (the complete live board, one row per market, no limit), ranks locally on returned fields, and reframes `featured` as one editorial highlight with a pointer to the board.

## What changed

- **Live signal board**: explicit board workflow; the catalog demo now leads with `./scripts/quotient.sh signals`; the `window` default-24 doc error is fixed (opt-in, no default).
- **First-use x402 budget**: on the first paid call with no autopay policy, the exit-10 cost preview carries a `preauth_offer`; the agent now relays it once per session — pre-authorize $1.00 of reads (caps per-call $0.05 / per-run $0.25 / per-day $1.00, spend ledger) — and writes a policy only on an explicit user yes naming the amount. Decline ⇒ per-request `--approve <token>` approvals, unchanged.
- **Current surface**: assets/asset-search/stance, market search, `/signals/perps` price outlooks across covered commodities (wti, gold, silver, copper, natural-gas, platinum), crypto (btc, eth), and single-name equities, `/latest`, X research/profile, forecast generation (`POST /api/auth/forecast-requests`, $1.00, async, prepaid-key only), and free availability/target resolvers. Removed surfaces (`/narratives`, `/signal-score`, `/signals/oil`) are gone from the docs.
- **Signal field semantics**: `basis_status: unresolved` + `grounding_status: unavailable` documented as the normal state for venue-resolved markets (no settlement feed claimed); the only signal availability gate is `venue_quote_unavailable`.
- **Multi-asset monitoring**: `pm.sh perps --symbol` and `pm.sh hl --coin` (defaults unchanged), plus the full Hyperliquid `xyz` coin map in `polymarket-monitoring.md`.
- **Doc posture**: the skill now describes what data exists and how to fetch it, and lets the agent reason — verbatim response templates, fixed section orders, and table ceilings removed. Neutrality (no buy/sell/hold), confirmation-first payments, execution gates, and the risk disclosure are unchanged.

## Review follow-through (#609)

All seven blocking items from the original review are enforced in the shipped scripts:

| Review item | Enforcement |
| --- | --- |
| Confirmation-first payments, autopay policy, caps, idempotent retries | `payments.sh` ships `confirm` mode; `autopay.json` caps + append-only spend ledger; per-URL memoization, single ledgered retry |
| Pin the x402 tuple; host allowlist | `payments.sh` preflights the full tuple: payee `0xC3d01FD2F79d4c57aD106AB8ecc12a5dE24F97cB`, Base USDC `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913` (or pinned RH-chain USDG), scheme `exact`, timeout ≤ 300s; hosts hardcoded, env can only narrow |
| No raw `EVM_PRIVATE_KEY` recommendation | `vanilla-x402-flow.md` requires a secret store / secure signer and pins package versions |
| Plan-bound `--execute` confirmation | Two-phase: `--execute` prints a hashed plan and exits 12; only `--execute --confirm <hash>` submits; stale/non-actionable/non-live-priced rows rejected |
| Outcome-aware book checks | `pm.sh price/book` resolve by outcome name (`--side` / `--outcome`) and verify `--expect-condition`; NO trades read the NO book |
| Verify execution, don't trust `completed` | `signal-strategy.mjs` verifies the mined receipt and resulting position; halts on Bankr scanner errors such as `untrusted_address` |
| Risk disclosure before execution | Embedded in every execution preview (`QP_RISK_DISCLOSURE`) |

## Generation

The folder is compiled from Quotient's canonical skill source by a build script and drift-guarded by tests (byte-identity, dual-auth coverage, venue/instrument coverage, and the new board/budget assertions). `contract-prices.json` is validated against the live OpenAPI (13.9.0).

https://claude.ai/code/session_01BGL6otjCPoJwdR2vLXpuWs


saltoriousSIG (changes requested):

- Trade confirmation is not bound to the action submitted to Bankr. `planHashOf()` omits the wallet, market slug/question, outcome, free-form `prompt`, and plan expiry, but `runConfirm()` later trusts and submits the unhashed `t.prompt`; changing the stored prompt preserves the approved hash and can execute a different trade. Please hash the complete immutable plan and submit a structured order/transaction whose chain, venue contracts, outcome token, side, amount, maximum price/slippage, fees, value, expiry, and transaction count are locally validated against that hash immediately before submission.
- The wallet used for planning is not authenticated before execution. Holdings and fill verification use user-supplied `--wallet`, while `/agent/prompt` spends from the wallet attached to `BANKR_API_KEY`; the code checks only after submission and incorrectly reads `walletAddress` even though `/wallet/me` returns addresses in `wallets[]`. Resolve the authenticated EVM wallet before preview or submit, require an exact match with `--wallet`, bind it into the plan hash, and abort if it is absent or different.
- Quote and success verification do not establish the confirmed fill. All trades are re-quoted before a sequential batch rather than immediately before each submission, the natural-language Bankr prompt has no enforceable limit price, and `verifyJob()` accepts any successful Base/Polygon receipt hash found anywhere in the job plus any positive position increase. Please re-quote per order, use an enforceable limit/max-price and expiry, and verify the structured job transaction hash on the exact Polygon target with the expected sender, market/outcome token, events, fees, and position delta consistent with the filled amount; otherwise report submitted-unverified and stop.
- The x402 preflight validates one matching challenge offer, but the subsequent opaque `bankr x402 call` fetches another challenge and may select a different offer; `--max-payment` does not pin network, asset, payee, facilitator, or timeout. Please bind payment to the exact validated challenge/offer and verify Bankr’s final `selectedRequirements` immediately before signing, including host, scheme, chain, token contract, payee, facilitator, amount, and expiry. Apply the same fix in `payments.sh` and `signal-strategy.mjs`.
- API-key host validation is bypassable with URL userinfo. The shell validators accept prefix-matching values such as `http://localhost:1234@evil.example`, whose actual destination is `evil.example`, and then send `QUOTIENT_API_KEY` there. Parse every configured origin, reject credentials/path/query/fragment, and require the parsed hostname to be exactly the pinned HTTPS host or literal `localhost`/`127.0.0.1` with a numeric port; validate `--add-host` before persisting it.
- A payment approval token is bound only to normalized route/count/price, not the approved command, full origin/path/query, method, or body. A token previewed for one filtered signals request can authorize a different request at the same route and price. Please hash and compare the full normalized request plan and command-relevant arguments before consuming the token in both payment implementations.
- Spend caps are not fail-safe under failures or concurrency. An ambiguous failed x402 call is automatically retried with a fresh payment while the first possible charge is ledgered as zero, and concurrent processes perform cap checks plus read-modify-write ledger updates without a shared reservation lock; both paths can exceed the user-approved run/day/total budget and lose ledger entries. Do not retry an ambiguous paid request unless the exact payment proof/idempotency identity is replayed, and reserve planned spend under a cross-process lock before payment, then atomically reconcile an append-only ledger afterward.
- The preferred CLI/update path bypasses catalog review and is incomplete when only this catalog skill is installed. `SKILL.md` tells agents to upgrade the CLI and run `quotient skill update auto` without user approval or an immutable source/version, while `quotient` is not a required installed binary and the vendored scripts do not implement the advertised paid forecast-generation POST. Please remove automatic updates, require explicit approval plus a pinned version/source and integrity for any CLI install/update, and vendor a safe generation command with an explicit $1 preview, body binding, persisted idempotency key/job ID, and same-origin polling.

PR #674, opened 2026-08-27
Add AgentRisk skill: pre-trade token risk scoring for Base

Neurobyteio:
Adds a pre-trade risk scoring skill for Base tokens — checks whether a token is a honeypot, how concentrated its holders are, whether its LP is locked, and whether its deployer wallet looks freshly created for a one-off launch. Combines direct on-chain checks, GoPlus Security, and DexScreener data, plus brand impersonation detection for tokens mimicking known companies.

Every response includes a scan_id, timestamp, and an explicit confidence flag — set to "low" instead of guessing when key deployer or liquidity data couldn't be verified. Paid per call via x402 (0.15 USDC), settled through Coinbase's official CDP facilitator — no API key needed.


Neurobyteio:
Hi - following up on this PR. It's been about 5 days and everything's passing checks. Let me know if there's anything I should adjust, or if it needs anything else from my side. Thanks for your time!

PR #676, opened 2026-08-28
feat: add Steer LP vault skill

derekbar90:
## Summary
- add a Steer LP vault skill with curated-vault safety boundaries
- add installed-CLI preflight and compact rebalance context
- add durable vault memory and manifest artifact guidance

## Validation
- `node --check steer/scripts/ensure-steer-cli.mjs`
- YAML frontmatter parse
- `git diff --check`


saltoriousSIG (changes requested):
High/medium issues I'd block on:

- The live and scheduled preflight automatically installs mutable npm `latest` globally without user approval or a repository-pinned artifact identity, then "verifies" whichever `steer` executable is first on `PATH` only by trusting its self-reported version. That executable receives the Steer secrets and is the authority that constructs every unsigned wallet transaction, so a compromised publisher/dist-tag, registry response, PATH entry, or concurrent global update can steal credentials or replace reviewed calldata while still reporting the expected semver. Please pin a reviewed CLI version and integrity/signing provenance in this repository, require explicit approval for installation/update, install it in an isolated immutable location with lifecycle scripts disabled, invoke that verified binary by absolute path, and make upgrades reviewed repository changes rather than automatic `latest` changes during transaction workflows.
- The submission boundary checks `request.transaction` against the same CLI preparation that supplied it, but never independently establishes which contracts and selectors are allowed or decodes the final calldata back to the user's intent. Pool metadata and successful simulation do not prove that an approval spender, vault, factory, manager, or implementation is an approved deployment; a buggy or malicious CLI can return internally consistent arbitrary `to`/`data` and the skill instructs Bankr to submit it unchanged. Please pin the reviewed chain deployments, factories/registries, proxy implementations and runtime hashes, allowed selectors/ABIs, and spender graph; verify live code and registry/pool/vault wiring before every write; and independently decode the final approval/create/deposit/tend transaction to validate target, selector, token, spender/recipient, amounts/minimums, ranges, value, and signer before confirmation.
- The `volumeUsd` fallback treats `steer subgraphs` output as a trusted credential destination: it inserts `STEER_SUBGRAPH_STUDIO_KEY` into any returned URL and calls `fetch` without validating the scheme, origin, path shape, or redirects. A malicious or misconfigured CLI can therefore exfiltrate the key to `https://attacker.example/api//...`. The advertised redaction also misses secrets embedded in URL paths and structured upstream errors bypass it entirely. Please use an exact reviewed HTTPS origin/path allowlist, reject redirects and malformed endpoint templates, avoid constructing credentialed URLs from executable output, and apply one conservative sanitizer to every emitted CLI/GraphQL error without forwarding raw URLs or bodies.
- `rebalance-context.mjs` is not fail-closed against the CLI it installs. Current CLI `4.5.3` writes `Total API calls: 1` to stdout before the `markets history` JSON, while `parseCliJson` passes the entire stdout to `JSON.parse`, so a normal successful history call cannot be parsed and the compatibility fallback is not selected. Separately, `buildContext` can return `status: "complete"` without matching the returned vault address, chain, or protocol to the requested values; it accepts case-insensitively equal non-address pool strings, treats missing chain/protocol/token identity as complete, and converts `null` numeric fields such as token decimals to zero. Please parse a guaranteed clean machine-output channel, strictly validate the full requested vault/chain/protocol/pool/token tuple and EVM address formats, preserve missing numeric data as missing, and add adversarial fixture tests proving malformed, noisy, missing, or mismatched data can never produce `complete`.
- Several reviewed lifecycle commands do not match the current CLI schema. Both tend examples encode two ranges as one comma-separated `--position`, which `4.5.3` rejects with `VALIDATION_ERROR`; each range must be a repeated `--position=<lower>:<upper>:<weight>`. The generic pool-create recipe always passes `--fee-tier`, but Aerodrome rejects that flag and requires `--tick-spacing`, while V4 additionally requires `--tick-spacing` and `--hooks`. Routine verification also mandates `--trace` even though setup does not require a `debug_traceTransaction`/`callTracer` RPC. Please provide tested protocol-specific command forms, use default receipt verification for completion and trace only after a capability check, and add executable smoke tests for every documented prepare/verify recipe.
- The unattended tend duplicate guard is a non-atomic check followed much later by preparation and submission, with the transaction hash recorded only after Bankr returns. Two overlapping scheduler runs can both observe no completed record and submit the same one-run authorization independently. Please atomically claim the durable run identifier before quote/preparation, bind the claim owner to the reviewed preparation hash, permit only that owner to submit, and reconcile stale or uncertain claims against Bankr/onchain state instead of allowing a second execution.

PR #685, opened 2026-09-01
voidly-pay: prepare encrypted hires and verify payment evidence

EmperorMew:
Adds **voidly-pay**, a prepare-and-verify skill for encrypted Voidly hires: seal the brief locally, validate payment signing/submission data, check receipt evidence across independent Base RPC operators, and verify returned artifacts. Bankr retains the wallet and payment execution. These helpers do not sign or broadcast payments or complete an end-to-end Bankr hire.

Current review head: **`a78b021d3ca5f6fbfe4ea3b348c497d0a31973b7`**.

The audit follow-up addresses six findings:

- Validate the complete grant before preview/signing, retain immutable lane/amount/hash context, reject stale or changed terms, and forward only the frozen signing/submission payloads returned by the local gates.
- Re-derive and verify Ed25519 signing identities; require private hirer files.
- Handle short writes, verify persisted bytes and permissions, and check synchronization/close results before reporting success.
- Cap actual descriptor reads and refuse observed file replacement, growth, symlinks and non-regular inputs.
- Require pinned discovery instead of following an unsigned index URL. State the remaining manifest-freshness limitation.
- Label receipt results as RPC-quorum inclusion and latest-head depth. Safe/finalized status is explicitly not checked.

Also corrected the Bankr policy guidance, exact-version installation instructions, grant-band scope, ciphertext exposure timing and signing-versus-broadcast gas claims. All changes are confined to `voidly-pay/`; dependencies and the existing lockfile are unchanged by this follow-up.

**Verification of this head:**

- Clean Git archive and fresh locked installation on Node **24.19.0**: **276 passed, 0 failed, 0 skipped**.
- Regression coverage includes the eight malformed-grant audit cases, both locked SDK payment lanes, callback mutation/accessors, expiry, file races, private permissions and short writes.
- Independent source and packaging reviews found no remaining material blocker in the documented prepare-and-verify scope.
- Registry audit reported zero known advisories at the check.
- Live pinned discovery passed. The catalog's disclosed historical first-party receipt returned `PROVEN` with two operators; this was a read-only recheck, not a new payment or third-party adoption claim.

Reproduce the local suite after approving the locked dependency installation:

```bash
cd voidly-pay
npm ci --ignore-scripts --no-audit --no-fund
npm test
node scripts/discover.mjs
```

The catalog also includes a dependency-free historical receipt check that can run before npm installation. Install the whole skill folder: the settlement verifier needs its two bundled Node-only helpers.

The catalog slug, folder, logo and install target align with the repository's catalog requirements. The `main/voidly-pay` install URL is the post-merge target. Bankr-host installation and an end-to-end Bankr-to-provider round trip have not been demonstrated. Local file protections depend on supported filesystem primitives; the reported runtime test was on macOS, not a cross-platform certification.

Ready for maintainer re-review and the Bankr merge/publication step. No approval, merge or live marketplace availability is claimed.

Contributed under Apache-2.0; Voidly retains copyright.


saltoriousSIG (changes requested):

- The documented EIP-3009 payment handoff does not require a complete final signing/transaction preview. Before either authorization is signed or Lane B calls `payForGrant`, require the Base chain ID, USDC contract, amount in both USDC and atomic units, payer, payee/spender, EIP-712 domain and full typed message, nonce/grant binding, validity window, and the exact submission transaction (USDC contract, selector/calldata, value, and estimated gas/fees). Reject stale or changed authorization terms. The current "exact amount (atomic USDC), payee, and grant hash" is not enough to safely approve a payment.
- Lane B explicitly creates a bearer `transfer` authorization that anyone holding the bytes can redeem, but the skill does not require local-only handling or forbid relaying that signature to the provider, facilitator, chat, logs, or other untrusted services. Please make this a hard refusal: keep the authorization in the wallet/local trusted submitter, pin and validate the recipient call, and require a fresh user confirmation if it ever must leave that boundary.
- `verify-settlement.mjs` accepts any number of `--rpc` values, including one arbitrary URL, then reports "quorum" and takes the confirmation height from only the first RPC. Require at least two independent, HTTPS allowlisted Base RPCs, validate each receipt's chain/block context, and compare the head/finality result across the quorum. Otherwise a single attacker-controlled endpoint can produce a false PROVEN result.
- The catalog and SKILL instruct users to run `npm install` for an untrusted third-party package before any verification, but ship no lockfile or integrity pin and leave both TweetNaCl dependencies on mutable caret ranges. Please require explicit user approval before installation and commit a reviewed lockfile with exact resolved versions and integrity hashes, or otherwise pin the package source and artifact before running third-party install scripts.

EmperorMew:
Updated the folder at fork commit `0eceb26` (the PR head). Every claim below is measurable in that folder: `npm ci --ignore-scripts && npm test` (213/213 on Node 20, 22, 24) and the catalog demo's dependency-free proof (`PROVEN`, exit 0, `quorum: 2/2 agreed`).

Thanks — all four are real, and #3 was live rather than theoretical. In order
of severity, then what else changed while closing them.

**#3 — a single RPC could produce PROVEN.** Reproduced before touching
anything: one `--rpc` printed `quorum: mainnet.base.org — receipts
byte-identical` and exited 0. `scripts/verify-settlement.mjs` now settles the
quorum before a packet leaves — at least two operators counted by **host**
(the same endpoint named twice, on another port, or as `localhost` beside
`127.0.0.1`, is one operator), HTTPS only, hosts from a reviewed
`ALLOWED_BASE_RPC_HOSTS` pin in `scripts/lib/pins.mjs`. Per operator:
`eth_chainId == 0x2105`, a byte-identical receipt, and `eth_getBlockByNumber`
at the receipt's height returning the receipt's own `blockHash`.
Confirmations come from the **lowest** head across the quorum, with a
30-block divergence ceiling. `--allow-unpinned-rpc` still exists as the
escape hatch for a paid archive node, but it no longer yields an ordinary
verdict: the first line becomes `PROVEN-UNPINNED` and the exit code is **2**,
so nothing that greps `PROVEN` or tests for exit 0 can mistake operators the
user chose for the reviewed allowlist. The transport refuses redirects, sends
`accept-encoding: identity`, and aborts the body at 4 MiB on the wire.

One narrowing on your scope list, since you were precise everywhere else: the
captured PROVEN output in SKILL.md was already produced against two
operators, so that example was not fabricated. What was wrong in it is that
the confirmations number came from `rpcs[0]` alone while the line above it
said "quorum". Corrected for that reason.

Beyond your finding, "payment proven for this exact hire" was itself an
overclaim: the nonce bound the transaction to the grant *hash*, but payer,
payee and amount were asserted exactly as typed. The verifier now takes
`--grant ./keep.grant.json` — the envelope `seal-hire.mjs` writes beside the
keep file — recomputes its hash byte-for-byte as the SDK's `envelopeHash`
(a test compares the two over random envelopes), reads payer and payee off the
grant — the `from`/`to` `buildTransferPaymentAuthorization` signs — and
requires the settled amount to sit inside the grant's price band, which is
what the SDK's own binding accepts. A typed value beside the grant that
disagrees refuses `grant_terms_mismatch`. Without `--grant`, the verdict now says in its own
`terms:` and `scope:` lines that the terms were typed and the grant's terms
were not checked.

**#1 — the preview was three fields.** It is now every field the EIP-712
signature commits to — chain and USDC contract, amount in atomic units *and*
decimal USDC, payer, payee off the verified manifest, the EIP-712 domain
(`USD Coin` / `2` / `8453` / `0x8335…2913`), the complete typed message, the
nonce beside the grant hash it derives from, the validity window, and on
Lane B the submission transaction — plus the rules around it: the signing
wallet is resolved first (`bankr wallet`) and must equal the grant's
`price_payer_account` or the hire is re-sealed rather than the preview
"fixed"; the manifest is re-verified immediately before signing and drifted
terms refuse; nothing prepared earlier in the session is ever signed. The two
Wallet API calls the SDK callbacks map to are named exactly (`/wallet/sign`
with `eth_signTypedData_v4` and the SDK's `typedData` unmodified;
`/wallet/submit` with `to`/`chainId`/`value: "0"`/`data`), and Lane B carries
your later reviews' transaction gates: exactly one transaction, locally
decoded before submit — `chainId 8453`, `to == USDC`, `value == 0`, selector
`0xe3ee160e`, ABI arguments equal to the approved typed message, no extra
calls — a scanner refusal such as `untrusted_address` stops the flow and is
never routed around, and a payment is reported settled only when
`verify-settlement.mjs --tx <hash> --grant ./keep.grant.json` prints
`PROVEN` and exits 0.

**#2 — the bearer authorization had no boundary.** Lane B's `transfer`
authorization is named bearer material with a hard local-only rule: never to
the provider's accept URL, a facilitator, a relayer, chat, tool output, logs,
or a file that is not `0600`; target contract, function, recipient and amount
pinned and checked before broadcast; if those bytes ever must cross that
line, a fresh human confirmation naming the destination. The phrase "or your
facilitator" — which authorized exactly the handoff the rest of the section
forbids — is deleted, and Lane A is stated as the default for this reason.

**#4 — install before verification, with no pin.** `package-lock.json` ships
(three packages, each with `integrity` and `resolved`), `package.json`
carries exact versions instead of caret ranges, and the instruction is
`npm ci --ignore-scripts` behind an explicit human go-ahead naming the three
packages and the registry. The catalog demo is reordered so the settlement
proof — which needs no npm package, only Node and `scripts/lib/pins.mjs`
beside it — runs *before* anything is installed.

One measured correction there, because it changes the reason the exact pins
matter: `npm ci` does **not** refuse `^1.0.3` against a lockfile resolved to
1.0.3 — that exits 0. It only errors `EUSAGE` when the lock cannot satisfy
`package.json` (checked against 1.0.2, exit 1). So the caret ranges were a
drift risk the next time the lock is regenerated, not a live hole under
`npm ci`. The exact versions are still the fix; I would rather state why
correctly.

**What else changed, because your review made us look harder.** Six
independent adversarial rounds were run over the folder after your four were
closed (an engine fuzzer with a hostile local operator, a hostile
index/manifest/registry harness, a doc-truth sweep that runs every captured
command, a supply-chain and Node-compatibility sweep, and a simulated
re-review against your own prior standard), and every finding is closed with
a regression test. The ones worth naming: a symlinked skill directory made
two of the CLIs exit 0 having done nothing; receipt verification enforced no
provider pin; the body cap was compared after gzip inflation; the payee
account and the price band were still read off the signed manifest (both are
pins now — a manifest carries no freshness or revocation, so a replayed
signed one could name any payee); and the first version of the Lane B gate
program estimated gas by sending the signed `transfer` calldata — the bearer
authorization itself — to public RPC operators. That last one is exactly the
handoff your finding #2 forbids, caught by our own round, not shipped to you.
`preview-payment.mjs` now never transmits signature bytes: its fee line is
typical gas × the live gas price, and it says so.

The Lane B gates are a program, not a paragraph: `scripts/preview-payment.mjs`
renders every EIP-712 field from the grant, decodes the SDK's request offline
(selector `0xe3ee160e` and all nine arguments) inside the `broadcast` callback
with the previewed amount binding by default, binds the `/wallet/sign` and
`/wallet/submit` responses' `signer` to the grant's payer, and accepts a
submission only when it is mined and successful. Node 20 is the floor and the
folder says so (`engines`, `.npmrc engine-strict`, a runtime guard) — on
Node 18 the SDK's hashing throws for lack of WebCrypto in ESM.

The frontmatter now says what the skill is: **prepare and verify** — seal
the wire, prove the settlement, check the artifacts. Submitting the wire,
paying, and opening the result are `@voidly/session` calls it deliberately
does not wrap, and no third-party round trip through Bankr has happened yet.
That sentence is in the file rather than implied by its absence.

Worth re-checking, roughly in this order:

- The quorum floor counts hosts, not URLs — two spellings, ports, or
  loopback aliases of one operator should still refuse.
- Whether `PROVEN-UNPINNED` / exit 2 is the right shape for the escape hatch,
  or whether a marketplace skill should not ship one at all.
- The `--grant` binding: that the dependency-free hash really is the SDK's
  (`tests/verify-settlement.test.mjs`, the envelopeHash test), that
  payer/payee are read from `price_payer_account` / `price_payee_account`,
  and that the band check (`price_min_amount..price_max_amount`) matches
  what the SDK's `bindAuthorizationToGrant` accepts.
- The preview and the Lane B gates in #1 are doc-enforced. There is no
  runtime on our side that can compel a Bankr agent to decode a transaction;
  the settlement proof is the after-the-fact check, and I have not claimed
  more than that.
- The allowlist has four operators; one (`base-rpc.publicnode.com`) is not
  archive and fails closed on older receipts. Four may be the wrong trade
  against pinning fewer and failing more often.

EmperorMew:
@saltoriousSIG all four points from your review are addressed at `0eceb26` — tests pass 213/213 on Node 20, 22 and 24, and the catalog demo's dependency-free proof returns `PROVEN` with `quorum: 2/2 agreed`. Happy to walk through any of it whenever you have a moment.

EmperorMew:
Head moved to `b79657d` — one thing changed, and it is the demo you would run first.

**What was wrong.** `verify-settlement.mjs` with no `--rpc` read `mainnet.base.org` + `base.gateway.tenderly.co`. On 2026-09-03 three consecutive runs refused `rpc_unanswered — mainnet.base.org: http 503`; a 12-call burst against that host returned `200 200 200 200 503 200 200 200 503 503 200 200`. The quorum floor is 2 distinct operators, so one rate-limited member of a two-member default pair meant the catalog's `PROVEN (exit 0)` could not be reproduced.

**What changed.**
- Default pair is now `base.gateway.tenderly.co` + `base-mainnet.public.blastapi.io` (two companies). Measured 12/12 and 10/10 archive reads of the example receipt under the same burst.
- `base-mainnet.public.blastapi.io` and `base-pokt.nodies.app` join the reviewed allowlist in `scripts/lib/pins.mjs` with those measurements. `mainnet.base.org` stays allowlisted, not default, with the 503 note.
- `httpRpc` retries HTTP 429/5xx and network-level fetch failures twice (400 ms, 1200 ms). Timeouts, the body caps, non-JSON bodies and JSON-RPC errors stay final on the first read. An operator that never answers still refuses the run — a retry can recover a dropped read, not manufacture a proof.
- `SKILL.md` and `references/settlement-proof.md` follow the pair; the `PROVEN-UNPINNED` example is a fresh capture.

**Reproduce from the folder, no install:**
```
node scripts/verify-settlement.mjs \
  --tx 0xb1ac733095c19e2e4829a3d448a02b8297d08e55f98678adfcba2e3e92747a3a \
  --grant-hash 5e63f8c4f11b989bac73b4306bb1a7975b91571a586989127b35f812c31daea6 \
  --payer 0x5cad296e06a976886a5d5bef831520c3d5965af0 \
  --payee 0xb0b3fca940e04f99367f08e665e1c2cb4ebd4912 \
  --amount 50000
# PROVEN … quorum: 2/2 agreed — base.gateway.tenderly.co + base-mainnet.public.blastapi.io, receipts byte-identical   (exit 0)
npm ci --ignore-scripts && npm test    # 213/213
```
Nothing else in the folder changed, and the four points from the review stand as addressed at `0eceb26`.

saltoriousSIG:
the scope of the PR changed materially so did one more review pass and these are the issues that would just need to be adjusted.

High issues I’d block on:

- None found in this pass.

EmperorMew:
Pushed a scoped follow-up in `11a1408914e21e79777814db5c57cd643089b438` for re-review.

- `check-sign-response` now requires an explicit lane and cryptographically recovers the grant payer over the exact EIP-712 domain/message. It no longer trusts only signature shape and the response's signer label.
- Lane B `check-request` requires the approved signing response and binds the calldata's `v/r/s` to that verified signature before any gas-price lookup. Signed bytes are not printed or sent to an RPC simulator.
- Signature/request files must be private bounded regular files. Descriptor checks, no-follow/nonblocking acquisition, and exact signed-second expiry have regression tests.
- Corrected the Bankr instructions against current official [sign](https://docs.bankr.bot/wallet-api/sign/), [submit](https://docs.bankr.bot/wallet-api/submit/), and [Terminal](https://docs.bankr.bot/security/bankr-terminal/) documentation: recipient restrictions block typed-data signing and raw submission; arbitrary calls default on, timers are optional, and the actual effective policy must permit the operation. Restrictions are never relaxed or routed around by this skill.
- Pinned `ethers@6.17.0` for local recovery. The settlement verifier remains unchanged and dependency-free. The standard test command includes compatibility checks against the actual locked `@voidly/session@1.0.0` builders and calldata.

Verification on Node24 in a fresh outside-repository copy with an empty environment/config and fresh npm cache:
```text
npm ci --ignore-scripts --no-audit --no-fund
added 12 packages in 823ms
npm test
tests 228; pass 228; fail 0; cancelled 0; skipped 0; todo 0
duration_ms 8336.8405
```
Independent source review also passed 228/228. Registry audit reported zero known advisories for the locked dependency tree at this check; that is not a security certification.

Keyless public discovery passed. The existing disclosed first-party settlement example still returned `PROVEN`, two independent operators with byte-identical receipts. No new payment, Bankr wallet/API signing, or delivery occurred. The skill remains **PREPARE and VERIFY, not a completed end-to-end Bankr hire**. Requesting current-head re-review; not claiming approval or merge.

justinquidli:
🎉 Thanks @EmperorMew! The payout agent sent you **3200 BNKR** on Base for this contribution.

> Large voidly-pay skill with security-audit follow-ups and 276 passing regression tests covering malformed grants, file races and payment lanes.

Tx: https://basescan.org/tx/0xa559a0f4d2c70445b49ec9207c5ffeb565d74b4a3e4a33a672cda4dcd7ea0cb1

Paid to your GitHub username via [Quidli Connect](https://connect.quid.li).

justinquidli:
🎉 Thanks @EmperorMew! The payout agent sent you **3800 BNKR** on Base for this contribution.

> Substantial new prepare-and-verify skill with six audit findings fixed and 276 passing regression tests.

Tx: https://basescan.org/tx/0x66d445a7652ca92cd0ef9ce45860a2d839477f7b9611a0671e85f88cdca25f6a

Paid to your GitHub username via [Quidli Connect](https://connect.quid.li).

PR #699, opened 2026-09-08
Add lienfi — research and buy tokenized US tax lien certificates on Base (USDC)

jackdishman:
## Summary

Adds **lienfi** — research and buy tokenized US tax lien certificates and redeemable tax deeds on [LienFi](https://app.lienfi.com), settled in USDC on Base (chain 8453).

The skill teaches the agent to:

- **Research with no credential** over LienFi's hosted MCP endpoint (`POST https://api.lienfi.com/api/v1/mcp`, JSON-RPC 2.0, nothing to install, no session): `market_overview`, `search_liens` (ranked on NET yield, after LienFi's share of the gain, with the same arithmetic the site displays) and `get_lien`.
- **Get authorized once, by a human.** The operator signs an authorization in a browser at `app.lienfi.com/agents/authorize` and hands the agent what that page prints; the agent signs a key proof and the consent agreements with its own wallet and registers. The skill has the agent check whether it is already registered before asking, have a stale authorization revoked first (a second one does not replace the first), and receive the bearer privately — never in a public X reply.
- **Buy from the wallet it controls**: `prepare_purchase` → sign the acknowledgment → `confirm_purchase` → submit `approve`, `buyNFT` and `setApprovalForAll` from its own wallet → `report_purchase` → `my_positions`. LienFi signs only the price quote; it never holds a key and never submits a transaction.

Both ways of holding a Bankr wallet are covered: Bankr's own agent (terminal, X, Telegram) signs and submits with its native wallet tools and needs no Wallet API key; an agent running elsewhere uses `/wallet/sign` and `/wallet/submit` with a key that has `readOnly` off and `allowedRecipients` empty.

## Contents

- `SKILL.md` — prerequisites, the money conventions that are not guessable from field names, six steps with a copy-ready request body for each, hard rules, refusals, what is verified and what is not, troubleshooting.
- `catalog.json` — `slug: lienfi`, demo prompts, setup steps, `install.type: bankr`.
- `logo.svg`
- `references/refusal-codes.md` — every `error.code` the server can answer and what to do about it, generated from the server's own table.
- `references/typed-data.md` — the EIP-712 envelopes the agent signs, field by field.

## How it was tested

[...]

PR #702, opened 2026-09-11
lienfi: point doc links at /docs, and scope the operator bearer to the conversation it was pasted into

jackdishman:
Two follow-ups to #699, both carried from the LienFi monorepo where `lienfi/` is
maintained. Docs-only — no tool calls, request bodies, refusal codes or contract
addresses change.

### 1. The API reference moved to `/docs`

It was at `app.lienfi.com/docs/api`; it is now `app.lienfi.com/docs`. Six URLs in
this folder follow it, including the two deep links `SKILL.md` tells the agent to
read *before its first call* — `#mcp-walkthrough` (a request body for every step)
and `#mcp-refusals` (every refusal code with what to do).

Verified live today: `/docs` serves 200 with both `mcp-walkthrough` and
`mcp-refusals` anchors present in the HTML, and `/docs/api` 308-redirects to it
permanently. So the merged copy keeps working either way — this removes the hop
rather than fixing a break.

### 2. The operator bearer is scoped to the conversation it was pasted into

Corrected against what was measured on 2026-09-08 with Bankr's own agent. It
carried the operator's bearer into a later call **in the same terminal
conversation** unaided — but a few hours later, asked on X, it held nothing for
the same wallet: it checked the wallet's status, found it registered, and asked
for the block again.

The skill previously read that state — `registered: true` while the agent holds
no bearer — as the authorization being lost, and sent the operator to **Revoke**
and sign again. That is the wrong first move and an expensive one, because a
second authorization does not replace the first; it is refused
(`registration_refused`). The page prints the blob once and does not store it,
but the operator pasted it somewhere, and it is still in that conversation and
valid until `expiresAt`. So the skill now asks for that block first, privately,
and treats re-registering with it as idempotent; only if it is genuinely gone
does anyone revoke and re-sign.

### On the review left on #699

@saltoriousSIG — your `CHANGES_REQUESTED` review is noted and still open. **This
PR is not a response to it.** It predates the review as monorepo work and is
scoped to the two items above.

One overlap worth naming rather than leaving for you to find: your fourth point
asks that the credential move to an approved secret store rather than
chat/transcript or model memory. Section 2 above edits that same passage, and it
documents observed behaviour rather than endorsing it as the target state — the
skill still says never to print the bearer, never to send it anywhere but
`api.lienfi.com`, and to keep it wherever the runtime keeps a credential. Happy
to fold this into a single response instead if you would rather review the
credential handling once, as a whole.

**Update (2026-09-11): closing this PR unmerged.** We've decided to leave the
published `lienfi/` as it was merged rather than carry these doc-link edits up.
Retracting the line that previously stood here saying a response to the full
review was being prepared — that is not currently planned, and I'd rather not
leave a commitment on the record that isn't going to be kept.

### DIFF
--- a/lienfi/SKILL.md
+++ b/lienfi/SKILL.md
@@ -79,8 +79,8 @@ address (see "Verified and unverified" at the end).
   (one lien, flat, carrying the NET per-year rate) and
   `https://api.lienfi.com/api/v1/liens?...` (the book with filters).
 - **Read before your first call** — https://app.lienfi.com/llms.txt (the money
-  conventions), https://app.lienfi.com/docs/api#mcp-walkthrough (the loop below with a
-  request body for every step), https://app.lienfi.com/docs/api#mcp-refusals (every
+  conventions), https://app.lienfi.com/docs#mcp-walkthrough (the loop below with a
+  request body for every step), https://app.lienfi.com/docs#mcp-refusals (every
   refusal code with what to do). Trust `tools/list` over any document, including this one.
 
 Keep the operator's bearer from the moment you have it, and send it on every wallet and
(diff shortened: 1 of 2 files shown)


jackdishman:
Closing this — we've decided to leave the published `lienfi/` as it was merged rather than following it with doc-link edits. The `/docs/api` links in the merged copy keep working: that redirect is permanent on our side and fragments survive it, so there's nothing for anyone to do here. Sorry for the noise, and thanks for merging #699.

PR #703, opened 2026-09-12
Sync the bankr skill with current platform capabilities

sidrisov:
Weekly sync of the `bankr` skill against the current state of the platform. Two groups below: what changed on the platform this week, and gaps the sync surfaced that predate it.

Docs-only — no behavior, no new files, and no change to the skill's frontmatter or `catalog.json`.

## Shipped this week

**LLM gateway — model registry**
- Added `gpt-6-astra` (latest frontier OpenAI) and `deepseek-v4.1-flash` (latest DeepSeek Flash, now vision-capable), and demoted the releases each supersedes so "latest" in the table means latest.
- Marked `gemini-3.5-flash-lite` and `gemini-3.1-flash-lite` deprecated, pointing at `gemini-3.8-flash`.

**LLM gateway — image generation**
- Image generation now runs on the `gpt-image-2.5` line: `-flare` for speed and high-volume work, `-sunburst` for precision, with `gpt-image-2` as the previous generation. Examples updated to `flare`.
- Documented that the gateway exposes `/v1/images/generations` but not `/v1/images/edits`, so Sunburst's editing advantage isn't reachable here; plus the `n` ≤ 4 cap and that the whole line is priced identically, so the choice is about output rather than cost.

**Max Mode**
- The skill claimed Max Mode accepts *any* gateway model. It doesn't: only the frontier, flagship and balanced lines are selectable. Light models stay reachable through the gateway API but aren't offered as the agent's model.
- Documented that the pickers list the two newest releases per line while the API still accepts older ids for as long as they're served — so dropping off a picker isn't the same as being gone. Removed XMTP from the surface list.

**Balance reads through the agent**

This is the one most likely to bite an agent consuming the skill. Agent balance tools now apply the wallet's low-value token preference and its $1 threshold, so a returned balance list is filtered by default and **is not the whole wallet**. Documented the `hiddenLowValueTokens` count, the `includeLowValueTokens` opt-out, and that native gas rows are never hidden — with the explicit warning that a token's absence from the list is not evidence the wallet doesn't hold it.

**Selling and transferring by ticker**

Ticker resolution for sells and transfers now runs against actual holdings with no USD floor, so a dust-sized position resolves like any other rather than falling through to a global search. Holdings form the candidate set rather than the answer, and ambiguity is resolved against the user's own contracts and balances.

**Token launch gates**
- A wallet whose only linked account is an email needs 72 hours of age rather than 24.
- Simulations are capped at 20 per wallet per 24 hours — still never consuming a launch slot, but no longer free to loop over.
- Gates apply on every launch path, not just the REST endpoints, and a blocked launch returns one generic message — so automation shouldn't branch on the reason.

**Agent profiles**
- Documented the `includeFeatured` query parameter on the list endpoint, including that the first page can then exceed `limit`.

## Gaps caught while syncing

These predate this week; the sync just surfaced them.

- **Per-IP deploy cap was missing entirely.** Non-partner deploys are capped at roughly 10 *successful* deploys per 24 hours per client IP, returning `429`. For anyone deploying programmatically from one host this is the ceiling they hit first — well before the per-wallet quota — so its absence was a real trap. Documented with the caveat that the figure is approximate.
- **x402 facilitator trust boundary was undocumented.** Verify and settle are authenticated, rate-limited and bound to the specific endpoint being paid for, and settlement pays the address on that endpoint's record rather than anything in the request. Added a note that the `facilitator` URL in a `402` response is informational and shouldn't be posted to directly.

## Verification

Checked the whole skill for stale references to anything retired this week — nothing left. Confirmed the pre-existing gaps above weren't recent regressions before labelling them as such. Reviewed the rendered Markdown for list and table breaks.

https://claude.ai/code/session_01Fh4pWnuc5PBsWWKPxidnxh

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fh4pWnuc5PBsWWKPxidnxh)_

### DIFF
--- a/bankr/SKILL.md
+++ b/bankr/SKILL.md
@@ -542,7 +542,7 @@ The [Bankr LLM Gateway](https://docs.bankr.bot/llm-gateway/overview) is a unifie
 - In OpenClaw config, prefix model IDs with `bankr/` (e.g. `bankr/claude-sonnet-5`). In direct API calls, use bare IDs (e.g. `claude-sonnet-5`). Run `bankr llm models` for the current model list
 - **Claude Code's `[1m]` context-tier suffix is optional through the gateway** — it's stripped before model lookup and the 1M window comes from the model's own context window, so `claude-opus-5` and `claude-opus-5[1m]` behave identically. If you do use it, quote it (`--model "claude-opus-5[1m]"`) — in zsh it's a glob character class and the command aborts before the CLI runs
 - **Per-model discounts** available for Bankr Club members and partners — applied automatically at billing time
-- **Image generation**: generate images via the OpenAI-native `/v1/images/generations` endpoint (model `gpt-image-2`), billed from the same LLM credit balance — see the reference
+- **Image generation**: generate images via the OpenAI-native `/v1/images/generations` endpoint — the `gpt-image-2.5` line (`-flare` for speed, `-sunburst` for precision) plus the previous `gpt-image-2`, all priced identically and billed from the same LLM credit balance — see the reference
 - **Expiring credit grants**: promotional or developer grants may carry an expiry date. Your spendable balance is your permanent (purchased) credits plus any unexpired grants — grants are spent first (soonest-expiring first) and drop off automatically at expiry
 - **Privacy tiers**: every request is served at `standard`, `zdr` (zero data retention), or `private` (TEE). Ask for a tier per request, per model, per base URL, or account-wide — see below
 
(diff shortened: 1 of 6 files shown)


justinquidli:
🎉 Thanks @sidrisov! The payout agent sent you **4500 BNKR** on Base for this contribution.

> Six merged PRs correcting materially wrong login/MFA, quota, model-registry and balance-filtering documentation in the bankr skill, preventing real agent failures.

Tx: https://basescan.org/tx/0xa1ac59f5b99ed8794c370764785b768b5ee07b5e4239a83f7f79cfe9e2ccabc0

Paid to your GitHub username via [Quidli Connect](https://connect.quid.li).

justinquidli:
🎉 Thanks @sidrisov! The payout agent sent you **2800 BNKR** on Base for this contribution.

> Six PRs correcting factually wrong platform documentation (launch quotas, key defaults, MFA login) that would have misled agents.

Tx: https://basescan.org/tx/0x4b9e98c3230d32b674b31b518df4c10268603d64a329d98678c3e342aebac5f1

Paid to your GitHub username via [Quidli Connect](https://connect.quid.li).

justinquidli:
🎉 Thanks @sidrisov! The payout agent sent you **1950 BNKR** on Base for this contribution.

> Combined payment for five accurate, high-value documentation-correctness PRs (#703, #708, #709, #719, #721, #727) that fixed misleading login/MFA/terms/API-key guidance agents would otherwise act on.

Tx: https://basescan.org/tx/0xc53d090d8e1da484f4b2e99e49c867bc5a2859e2e7a698412636eacdacc89033

Paid to your GitHub username via [Quidli Connect](https://connect.quid.li).

PR #720, opened 2026-09-18
Add askgrokwallet — spend guardrails and verifiable receipts for a Bankr agent

richard7463:
Adds the `askgrokwallet` skill: a policy gate and a receipt layer that sits **around** a Bankr agent's existing execution path. Nothing custodial — it never touches keys, never signs, never moves funds.

---

## Since your review (`64a0c64`)

All four required fixes are in. Two of them needed the server changed, not just the prose.

**1. P1 — evaluate the asset/counterparty, not just the venue.** Reproduced live before changing anything: `never trade PUMPDUMP` with `target: "uniswap"` and the asset in the summary returned `allow`. The engine now matches deny keywords against the request's whole stated identity — `target`, `assets`, `counterparty`, `recipient`, `payee`, `merchant`, `token`, `pair`, `market`, `summary` — and the reason names the field it matched:

```
deny rule "pumpdump" matched summary: swap 25 USDC for PUMPDUMP
deny rule "pumpdump" matched assets: PUMPDUMP
```

Two further bugs fell out of fixing it properly: the compiler treated the action verb as an asset (`never trade PUMPDUMP` → `["trade","pumpdump"]`), and the stopword stripping was substring-based, so `never pay coinbase` compiled to the keyword `base`. Both fixed, with assertions. Where a restriction cannot be expressed in these fields the API now answers `ask` — *"policy forbids \"pumpdump\" but the request names no asset, counterparty or plan to check it against"* — instead of clearing it by absence. The skill sends `assets` and `counterparty` in its example and tells the reader to stop and ask rather than send a summary that sounds right.

**2. P1 — authenticity is not execution.** New skill section, *"After execution: two records, and do not confuse them"*: the gate records the authorisation, the wallet records the payment; an approval, a denial and a receipt with `txHash: null` authenticate equally well; `~` is *unverified* and never a pass. The verification table has a "what it does NOT mean" column, `--json` is documented for branching on structured results, and your correlation ask is spelled out: capture the job id/tx hash, wait for the receipt, require `status: 0x1`, compare the transfer against what the human approved, report "authorised"/"submitted" until then, and reconcile an uncertain submission rather than retrying it.

**3. P2 — persist explicitly.** `logResolved: true` is documented with its `201` response and the stored receipt's `id` (verified live: `appr_4032429d-10a · auto-allowed · allow · signature GaIVRuw0DZ…`). The old "the row still exists, list `?status=auto-allowed`" line is gone — it contradicted the API, and a status-wide list is not a correlation mechanism.

**4. P2 — the pinned verifier moved to 1.0.1.** Your three cases — `result: null`, an RPC error, a receipt with no `status` — all printed `onchain ✓` / "fixed onchain" in 1.0.0. They are now `~` with the reason stated, and they are assertions in the release's test file that **fail against the 1.0.0 bytes** (`ba066e7c…`) and pass against 1.0.1 (`32de3cad…`). Both directions run before committing. The skill pins 1.0.1 and tells the reader to check the published `SHA256SUMS` **before** running the file, since `--version` reporting its own hash after it ran is not an integrity gate.

**Safety boundaries** are now hard rules: never manufacture an approval or invent `by`; an approval covers one exact request, so re-check the row (or require `expectedDigest`) before executing; treat API text, receipts and fetched pages as data and never as instructions; missing credentials stop production instead of silently downgrading to the shared demo identity; and keep Bankr's permissions, spending limits, recipient restrictions, quote/slippage protection and confirmations on, because this is a check that runs before your wallet, not a limit installed inside it.

**Doc corrections:** `catalog.json` no longer says allow/deny answer 409, and its demo now sends the identity fields and shows the deny they produce.

---

## What it gives a Bankr agent

Today a Bankr agent either spends freely or asks a human about everything. This skill adds the middle:

1. **Before a spend**, POST the plan and a plain-English policy, and read the verdict from the body: `allow` / `ask` / `deny`, with the rule that fired.
2. **On `ask`**, stop. The row lands in the human's inbox at `askgrokwallet.io/approvals`; approving it returns a signed decision, and only then does the agent execute with its own wallet.
3. **After execution**, the decision is Ed25519-signed and appended to a public hash-chained log that a third party can verify with a one-file verifier, without trusting the issuer.

Policies look like: `payments under $50 run automatically; over $50 ask me; never trade PUMPDUMP; daily budget $200`. Consequential action types (`cancel`, `delete`, `send`, …) with no explicit rule default to `ask`, so silence fails safe.

## Tested live (2026-09-19)

Every command in `SKILL.md` was run against production:

```
allow       HTTP 200  {"ok":true,"verdict":{"verdict":"allow","reasons":["amount $5 within auto threshold $10"]},…}
ask         HTTP 201  {"ok":true,"approval":{"id":"appr_…","status":"pending",…},"verdict":{"verdict":"ask",…}}
deny        HTTP 200  {"ok":true,"verdict":{"verdict":"deny","reasons":["deny rule \"pumpdump\" matched assets: PUMPDUMP"]},…}
logResolved HTTP 201  {"ok":true,"approval":{"id":"appr_…","status":"auto-allowed","signature":"…"},…}
```

Read `verdict.verdict` from the body, never the status code — `allow` and `deny` are answers, not failures, and the body is the contract that survives both this version and the previous one.

Also verified: `POST /api/approvals/{id}` returns `status: "approved"` plus a fresh signature; `GET /api/presets` returns five starting policies; a `never X` rule matches the whole stated identity of the request.

The verifier is a released, pinned artifact: `verify-receipt.mjs --version` prints `1.0.1` and the sha256 of the bytes you hold, `test-verify-receipt.mjs` checks the verifier itself offline, and the release ships `SHA256SUMS`. Two outside reviewers have broken it — a broadcast anchor read as "block 0", then an unknown anchor outcome read as "fixed onchain" — and both repros are now failing-against-the-old-bytes tests in that repo. That is why the verification step is in this skill at all, and why the skill says plainly that a four-tick receipt is a statement about the record, not proof that a payment happened.


saltoriousSIG (changes requested):
## Review: changes requested

Reviewed all six commits and the four-file diff at **`be4b652608745420269d05ad181c36ee6a118a8e`**, including the released verifier this skill instructs agents to execute. There are useful safeguards here—user-defined policies, explicit stops on `ask`/`deny`, no splitting to evade a denial, and a warning against production use of the shared demo identity—but the trade gate and receipt guarantees need correction before merging.

### Required fixes

**1. [P1 — high] Evaluate the traded asset/counterparty, not just the venue.**  
[`SKILL.md:41–45`](https://github.com/BankrBot/skills/blob/be4b652608745420269d05ad181c36ee6a118a8e/askgrokwallet/SKILL.md#L41-L45)

The example combines `target: "uniswap"` with `never trade PUMPDUMP`, but the engine only searches `target` for the prohibited token. I reproduced the following with synthetic, non-executing demo evaluations using the example policy:

| Summary / amount | Target | Actual verdict |
| --- | --- | --- |
| `swap 5 USDC for PUMPDUMP` / `$5` | `uniswap` | **allow** |
| Same request | `PUMPDUMP` | **deny** |

A blocked asset can therefore pass when an agent follows the venue-shaped example. Derive the evaluated identity from the actual planned transaction, including the canonical token/counterparty, and keep venue information separate. Stop for clarification when the API cannot represent a policy restriction faithfully. Bind the verdict to the exact chain, assets, recipient, amount, and execution bounds; changing those requires reevaluation.

**2. [P1 — high] Do not present receipt authenticity as proof of payment execution.**  
[`SKILL.md:86–100`](https://github.com/BankrBot/skills/blob/be4b652608745420269d05ad181c36ee6a118a8e/askgrokwallet/SKILL.md#L86-L100) and [113–114](https://github.com/BankrBot/skills/blob/be4b652608745420269d05ad181c36ee6a118a8e/askgrokwallet/SKILL.md#L113-L114)

The loop executes through Bankr independently, then promises a receipt of what happened, without a step that records Bankr's execution result. “The verifier printed four lines” is also insufficient: it prints those lines for failed or inconclusive checks, and its positive verdict authenticates a record rather than checking the underlying payment.

Using locally signed synthetic fixtures and stubbed HTTP/RPC, both an **approved** receipt and a **denied** receipt with `txHash: null` and `settledAt: null` produced four `✓` results and exit code `0`. That is valid evidence of an authentic approval/denial record; it is not a payment confirmation. The verifier's RPC calls checked the **log anchor**, not a payment transaction.

Document how to capture and correlate the actual Bankr job/transaction result. Before reporting a completed payment, verify its successful chain receipt and the relevant transfer/swap outcome against the authorized request. Describe approval receipts and execution receipts separately. Inspect structured verification results; `~` means the corresponding claim remains unverified.

**3. [P2 — medium] Persist auto-allowed receipts explicitly and retrieve the exact record.**  
[`SKILL.md:67–68`](https://github.com/BankrBot/skills/blob/be4b652608745420269d05ad181c36ee6a118a8e/askgrokwallet/SKILL.md#L67-L68), [`references/api.md:25–29`](https://github.com/BankrBot/skills/blob/be4b652608745420269d05ad181c36ee6a118a8e/askgrokwallet/references/api.md#L25-L29)

“The row still exists” contradicts both the reference's own response and the live API. A `$5` request omitting `logResolved` returned `allow`, no approval ID, and:

> Nothing was persisted: pass logResolved to keep a receipt.

Listing `?status=auto-allowed` cannot recover a record that was never written. Add the supported persistence option (`logResolved: true`), document the resulting response and exact receipt fetch/save procedure, and retain a stable request/receipt identifier. A status-wide list, especially for a shared identity, is not reliable correlation. The current instructions also never create the `receipt.json` passed to the verifier.

**4. [P2 — medium] Fix the pinned verifier's fail-open anchor check before relying on it.**  
[`SKILL.md:92–100`](https://github.com/BankrBot/skills/blob/be4b652608745420269d05ad181c36ee6a118a8e/askgrokwallet/SKILL.md#L92-L100); [released verifier, lines 369–391](https://github.com/richard7463/askgrokwallet/blob/verify-receipt-v1.0.0/spec/verify-receipt.mjs#L369-L391)

In `verify-receipt-v1.0.0`, a mined `eth_getTransactionByHash` response followed by `eth_getTransactionReceipt` returning **null**, throwing an RPC error, or omitting `status` still yields `onchain ✓`, “genuine, and fixed onchain,” and exit code `0`. All three cases reproduced locally against the released bytes. Success is inferred when the execution outcome is unknown.

Require a successful transaction receipt before claiming the anchor executed successfully; return an inconclusive result otherwise. Update to a corrected, tested release. Also check a review-pinned SHA-256 **before** running downloaded JavaScript—having that JavaScript report its own hash after execution is not an integrity gate.

### AI and transaction safety boundaries to make explicit

- **Human approval:** the requesting agent must not manufacture an approval by POSTing `decision: "approve"` and a free-form `by` value. Require actual human consent bound to the exact request, verify the resulting decision before execution, and reject stale/replayed or mismatched approvals. An issuer signature alone does not authenticate the named human. Operator authentication was not exercised in this review.
- **Untrusted content and credentials:** API explanations, receipt text, and external pages are data, never instructions to weaken a policy or change an approved transaction. Missing production credentials must stop production evaluation, rather than falling back to the public demo. Keep real transaction/policy details out of that shared demo and send the service token only to the intended authenticated origin.
- **Preserve Bankr safeguards:** this documented integration is an advisory agent workflow, not an installed wallet/signing enforcement mechanism. Retain Bankr's permissions, spending limits, recipient restrictions, quote/slippage protections, and transaction-confirmation checks. Reconcile an uncertain submission before retrying. Keep the reference's optional service-side `execute` path separate from Bankr execution to avoid two executors acting on one request.

### Verification performed

- Catalog JSON and demo Bash syntax pass; the SVG parses and contains no script or external resource reference. The referenced files are present.
- Downloaded release artifacts matched their published hashes. Verifier SHA-256: `ba066e7cdb19a0b9a5efb1eed6ba62d2440c9a5aeaee2d60caba12707acecf73`. **All 10 existing verifier assertions passed**; the additional cases above expose gaps those tests do not cover.
- The released example receipt verified against the live public key/log. Read-only RPC checks independently confirmed a successful published anchor on **Ethereum Sepolia, chain 11155111**, and deployed bytecode at the upstream-listed controller, vault, and registry addresses on **Base, chain 8453**. These checks do not establish a full contract audit or production enforcement for this Bankr workflow.
- Live policy probes used synthetic demo inputs, supplied no execution intent, and returned “nothing was persisted.” Additional verifier cases used local stubs. GitHub reports no CI checks for this PR.

**Small documentation corrections:** `catalog.json` still says `allow`/`deny` return 409 after the last two commits changed the other files to 200. Also, `references/api.md:61–63` says `target: "pumpdump-coin"` returns `allow`; the live probe returned **deny**, consistent with the documented substring matching.

richard7463:
Thanks — all four required fixes are in, and two of them needed the server changed, not just the docs. Pushed as `64a0c64`.

**1. P1 — evaluate the asset/counterparty, not just the venue.** You were right, and the live API reproduced it exactly as you described: `never trade PUMPDUMP` with `target: "uniswap"` and the asset in the summary returned `allow`. The engine now matches deny keywords against the request's whole stated identity (`target`, `assets`, `counterparty`, `recipient`, `payee`, `merchant`, `token`, `pair`, `market`, `summary`) and the reason names the field it matched:

```
deny rule "pumpdump" matched summary: swap 5 USDC for PUMPDUMP
deny rule "pumpdump" matched assets: PUMPDUMP
```

Two related fixes came out of doing this properly. The compiler was also treating the verb as an asset — `never trade PUMPDUMP` compiled to `["trade","pumpdump"]` — and the stopword stripping was substring-based, so `never pay coinbase` compiled to the keyword `base`. Both corrected, with assertions. And where a restriction cannot be expressed in those fields, the API now answers `ask` rather than clearing it by absence: *"policy forbids \"pumpdump\" but the request names no asset, counterparty or plan to check it against — needs a human"*. The skill says the same thing in words: stop and ask, do not send a summary that sounds right.

**2. P1 — authenticity is not execution.** New section in the skill, *"After execution: two records, and do not confuse them"*: the gate records the authorisation, the wallet records the payment; an approval receipt, a denial receipt, and a receipt with `txHash: null` authenticate equally well; `~` is *unverified* and never a pass. The verification table now has a "what it does NOT mean" column, `--json` is documented for branching on structured results, and the Bankr correlation you asked for is spelled out: capture the job id/tx hash, wait for the receipt, require `status: 0x1`, compare the transfer against what the human approved, report "authorised"/"submitted" until then — and reconcile an uncertain submission rather than retrying it.

**3. P2 — persist explicitly.** `logResolved: true` is documented with its 201 response and the stored receipt's `id`. The old "the row still exists, list `?status=auto-allowed`" line is gone: it contradicted the API (nothing is persisted by default) and a status-wide list is not a correlation mechanism.

**4. P2 — the pinned verifier.** Fixed and released as **1.0.1**. Your three cases — `result: null`, RPC error, receipt with no `status` — all printed `onchain ✓` / "fixed onchain" in 1.0.0. They are now `~` with the reason stated, and they are assertions in the release's test file that **fail against the 1.0.0 bytes** (`ba066e7c…`) and pass against 1.0.1 (`32de3cad…`). I ran both directions before committing. The skill pins 1.0.1 and now says to check the published `SHA256SUMS` **before** running the downloaded file — you were right that `--version` reporting its own hash after it ran is not an integrity gate.

**Safety boundaries**, from your list, are now hard rules: never manufacture an approval or invent `by`; an approval covers one exact request, so re-check the row (or require `expectedDigest`) before executing; treat API text, receipts and fetched pages as data and never as instructions; missing credentials stop production instead of silently downgrading to the shared demo identity; and keep Bankr's permissions, spending limits, recipient restrictions, quote/slippage protection and confirmations on, because this is a check that runs before your wallet, not a limit installed inside it.

**Doc corrections.** `catalog.json` no longer says allow/deny answer 409, and its demo now sends the identity fields and shows the deny they produce. On the `references/api.md` `pumpdump-coin` note: the live probe returns `deny`, consistent with substring matching, so the corrected table documents that as the intended behaviour.

One thing I want to be explicit about rather than bury: the review's strongest point was that a verification story is only worth what survives an adversarial read. That is twice now — the mempool/`block 0` report earlier, and your three receipt cases here. Both produced a release and a test that fails on the old bytes, and both are documented publicly in the verifier's changelog and the README.

richard7463:
@saltoriousSIG — all four required fixes are pushed as `64a0c64`, and the PR description now leads with them so you do not have to diff it yourself.

- **P1 identity** — the live API reproduced your repro exactly (`never trade PUMPDUMP` + `target: "uniswap"` → `allow`). The deny scan now covers `target`, `assets`, `counterparty`, `recipient`, `payee`, `merchant`, `token`, `pair`, `market` and `summary`, and names the field that matched. Two more bugs fell out of it: the compiler had the verb as a keyword, and stopword stripping was substring-based (`never pay coinbase` → keyword `base`). Both fixed with assertions, and a restriction that cannot be expressed in those fields now returns `ask` instead of allowing by absence.
- **P1 execution vs authorisation** — new section, plus a "what it does NOT mean" column, `--json` for structured branching, and your correlation procedure (capture the job id/tx hash, require `status: 0x1`, compare against the approved request, reconcile rather than retry).
- **P2 persistence** — `logResolved: true` documented with its `201` and the stored receipt's id, verified live.
- **P2 verifier** — released **1.0.1**. Your three cases (`result: null`, RPC error, no `status`) print `~` now; they are assertions that fail against the 1.0.0 bytes and pass against 1.0.1, and both directions were run before committing. The skill also says to check the published `SHA256SUMS` *before* running the file, per your note.

No format, canonicalization or signature rule changed, so nothing your review already verified needs re-verifying. If it is easier to review as a diff than as prose: <https://github.com/BankrBot/skills/pull/720/commits/64a0c64>

Happy to be told any of it is still wrong — the earlier `block 0` report and your three receipt cases are both public in that repo's changelog now, with the failing-on-old-bytes tests kept next to them.

PR #724, opened 2026-09-18
fix: Hunch Bazaar native X market creation without user API keys

rajkaria:
## Summary

Update the existing Hunch Bazaar listing to the immutable **3.1.0** release:
https://github.com/rajkaria/hunch-skills/tree/9b9be2e0cbe4a7c8dbf59fd0f05aaaca7c03bad4/hunch-bazaar

This fixes X market creation for each authenticated Bankr user, without asking users to configure an API key, wallet variable or requesting-user ID.

## Root cause and solution

The previous skill sent hosted creation through a standalone Wallet API adapter requiring credentials and durable local files. Bankr CLI sandboxes reset between invocations. The existing Hunch betting skill already delegates signing to Bankr; creation now uses the equivalent native personal-sign flow.

Bankr Terminal tool discovery reported the actual native tool as `sign_data` with `{payload, payloadType: "message", reason}`. The native bridge validates the full Bazaar challenge before signing; the requesting user's own Bankr wallet signs it. It does not use a shared Hunch or operator key.

Preview records use Bankr's permanent wallet-scoped `write_file` / `read_file` tools. The original bot reply commits to the exact normalized terms with a SHA-256 preview ID; confirmation is bound to the same authenticated wallet, handle, original X mention, body and five-minute expiry. Temporary sandbox files are not authorization state.

The flow is: request → exact preview + current creator fee → same-user confirmation → validated native signature → publish → live link. Original-post lookup, source-post uniqueness and server nonce verification protect retries. The live fee API currently returns a fixed 50% creator share of settlement fees, included in the approved body.

## Review scope

This PR changes only the `hunch-bazaar` catalog pointer and setup text. All executable source, tests and security notes are in the pinned release. The original `hunch` betting listing is unchanged. Standalone adapter payments and standing grants remain explicitly scoped to durable hosts; this release does not claim they work in ephemeral hosted sandboxes.

## Validation

- 46 script tests pass under Node and Bun, including independent wallets, tampered previews, expiry, challenge substitution, source ownership and dropped-response reconciliation.
- 602 focused Hunch routing, publication, wallet-proof and security tests pass.
- Production draft and unsigned create endpoints returned a valid preview and a challenge accepted by the exact-message validator. No market or financial transaction was created by that probe.
- Read-only Bankr Terminal diagnostics confirmed native signing inputs and permanent wallet file tools. A real confirmed X publication has not yet been performed with this release.

## Acceptance check

Install the pinned release in a Bankr account, mention `@bankrbot create a market on Hunch: <question, criteria and deadline>`, confirm the displayed preview, and verify the returned market belongs to that user's Bankr wallet. Repeat with a second account to verify account isolation. Neither account should be asked to set API credentials. This is the remaining end-to-end launch check.


saltoriousSIG (changes requested):
## Review: request changes

The release has substantial AI-agent and payment safeguards, and **all 46 published tests pass locally on Node 22.19.0**. I found one reproducible consent-expiry bug in the new native bridge that should be fixed before updating the catalog pin, plus a smaller recovery issue.

**Scope:** PR head `470d55b6f06282523b703072ac5e759fba0bbee1`, both PR commits, and the executable release it selects: `rajkaria/hunch-skills@9b9be2e0cbe4a7c8dbf59fd0f05aaaca7c03bad4`. I compared that release with the previously pinned `cabe9508…` and inspected the existing payment and webhook controls as well as the new creation flow.

### Must fix — P2: enforce preview expiry at the signing/publication boundary

**Location:** pinned [`scripts/bazaar-native.mjs`, lines 53–68](https://github.com/rajkaria/hunch-skills/blob/9b9be2e0cbe4a7c8dbf59fd0f05aaaca7c03bad4/hunch-bazaar/scripts/bazaar-native.mjs#L53-L68).

`readPreview()` checks the five-minute expiry once, before awaiting the original-post lookup. `prepare` also awaits the unsigned challenge. Neither branch rechecks preview expiry after those requests. `validateChallenge()` checks the challenge's own freshness, which is a separate clock and does not enforce the preview deadline.

**Reproduced with mocked HTTP and clock:** create a preview expiring at `12:05:00`; prepare a valid challenge at `12:04:59`; start `submit` and let its lookup take two seconds. The bridge sends the signed market-creation POST at **`12:05:01`**. A separate probe shows `prepare` returning a signing payload after a slow challenge response has crossed the same deadline. These are client-side reproductions; no real signature or market was created.

**Why it matters:** the agent is instructed to obtain fresh consent after expiry, but the deterministic bridge can proceed using expired approval. An otherwise fresh wallet proof does not repair that consent gap.

**Requested change:** recheck the committed preview's expiry immediately before returning `signData` and immediately before dispatching the signed POST, after all preceding network waits. Add regression cases that advance the clock during lookup/challenge retrieval and assert no signing payload/signed POST is emitted. Publish the fix under a new immutable commit and update this PR's pin.

### Smaller follow-up — P3: allow read-only recovery after expiry

**Location:** [`scripts/bazaar-native.mjs`, lines 17–23 and 53–54](https://github.com/rajkaria/hunch-skills/blob/9b9be2e0cbe4a7c8dbf59fd0f05aaaca7c03bad4/hunch-bazaar/scripts/bazaar-native.mjs#L17-L23).

`lookup` calls the same expiry-enforcing `readPreview()` before making any request. If publication succeeded but the response was lost, a recovery attempt after five minutes throws “show a new preview and obtain confirmation” without checking the existing market. I reproduced this with an already-created market in the mock server. The documented direct GET is a workaround, so this is a recovery/usability issue rather than evidence of duplicate creation.

Separate capsule integrity/identity validation from permission to perform a write: allow an expired, otherwise valid capsule to do the original-post lookup and creator check, while continuing to block expired signing/publication. Test delayed recovery explicitly.

### Safety controls confirmed in the source

| Area | Assessment |
| --- | --- |
| Prompt injection and tool use | Posts, market fields, API prose, and webhook content are explicitly untrusted. Origins are pinned, redirects are refused, native CLI transport is base64url, and webhook notifications discard supplied instructions and run read-only. |
| Consent and identity | Native previews commit to the body, wallet, handle, original post, and expiry. Instructions require same-user confirmation and the preview ID from the original bot reply. Capsule tampering and changed identities are tested. **The runtime provenance of those inputs still needs the integration check below.** |
| Wallet signing | Native creation uses only the exact validated EIP-191 `create_market` message: domain, action, chain, wallet, body hash, nonce, and timestamp are checked. This path does not request a token transfer or allowance. |
| On-chain payment bounds | The standalone betting adapter pins Base `8453`, Base USDC, the settlement recipient, and the exact approved amount. It builds EIP-3009 transfer authorizations with expiry at `now + 120s`; no approval/Permit2/unlimited allowance path is used. |
| Duplicate spending and automation | Durable payment records, locking, retained authorizations, and uncertain-signature blocking protect retries. Standing grants enforce local scope, amount, outcome, count/total budget, expiry, and revocation. The release explicitly excludes these durable-host flows from ephemeral hosted sandboxes. |

**Important on-chain boundary:** the settlement recipient is a Hunch-controlled account, not rule-enforcing market escrow. Bet recording, settlement, and refunds depend on Hunch's backend/operator. The existing disclosure should remain visible before stake/grant confirmation. A payment receipt proves a transfer, not successful bet recording, payout solvency, or an enforced 48-hour refund. This review does not establish those backend guarantees.

### Validation and launch acceptance

- Re-ran the complete public suite: **46/46 passed** on Node 22.19.0. The pinned release's GitHub security workflow is also green; this Bankr PR currently reports no checks of its own.
- Three additional isolated safety assertions exposed the two issues above: expiry during prepare, expiry during submit, and delayed read-only recovery.
- The PR's claimed 602 backend tests are outside this public suite and were not independently reproduced. Read-only Hunch API probes could not connect from this environment, so live service behavior was not verified.
- Before launch, attach the already-described **two-account end-to-end acceptance evidence**: actual X request → wallet-scoped file write/read → original bot preview → same-user confirmation → native signature → published market owned by that wallet. Include a cross-user confirmation refusal and a fresh-invocation readback. The current mocked tests exercise the bridge, not Bankr's actual identity, file-tool, signer, or X-thread integration.

**Bottom line:** fix the consent-expiry race and repin; then demonstrate the real two-account flow. The reviewed safeguards provide a good foundation, with the operator-custody dependency clearly disclosed.

rajkaria:
**Updated: the pin now points at [`rajkaria/hunch-skills@78803b0`](https://github.com/rajkaria/hunch-skills/tree/78803b097712d74657e2798e9fc0c68f6a3bd1c0/hunch-bazaar) (Bazaar 3.1.1).** The `hunch` betting listing pin is unchanged; its content is identical at the new commit.

### P2: expiry enforced at the signing and publication boundary

`scripts/bazaar-native.mjs`:

- Capsule integrity (`readPreview`: commitment hash, version, identity, body invariants, a well-formed `expiresAt`) is now separate from permission to write (`requireConsent`: `now < expiresAt ≤ now + 5 min`).
- `prepare` checks consent before requesting the challenge and again once it arrives, immediately before returning `signData`.
- `submit` checks consent after the original-post lookup, immediately before dispatching the signed POST, with no `await` in between.

`test/native.test.mjs` moves the injected clock while requests are in flight. The preview expires at 12:05:00:

- `prepare` at 12:04:59 with a 2 s lookup: refused, and no challenge is requested.
- `prepare` at 12:04:59 with a 2 s challenge response: refused, and no `signData` is returned.
- `submit` at 12:04:59 with a 2 s lookup (your repro): refused, and no signed POST is sent.

The exact boundary is pinned: at `expiresAt − 1 ms` the bridge signs, at `expiresAt` it refuses. Removing any one of the three checks fails its own test.

### P3: read-only recovery after expiry

- `lookup` needs integrity only, so an expired but intact capsule still runs the original-post lookup and creator check. After expiry it still refuses tampered or forged capsules, other identities, and a market owned by another creator.
- `prepare` and `submit` also reconcile the original post before checking expiry. A retry after a lost response returns the existing market instead of asking for a new preview. With no market, they refuse and never write.
- Tested: the create response is dropped, then `lookup`, `prepare` and `submit` run ten minutes after expiry. Each returns the market with no new signature or write. An expired `lookup` with no market returns `{found:false}`.

### Also changed

- A draft with an unreadable `confirmBy` now fails with a clear error instead of a canonical-JSON `TypeError`.
- SKILL.md step 8 now sends recovery through `bazaar-native.mjs lookup`.

### Validation

54/54 `npm test` on Node 22 in the release's GitHub Actions run ([35707088381](https://github.com/rajkaria/hunch-skills/actions/runs/35707088381)), and locally on Node 25.9 and Bun 1.3.11.

### Two-account acceptance

Not run yet: it needs two real Bankr accounts and live X posts. The evidence will be attached here before launch:

- X request → wallet-scoped file write/read → original bot preview → same-user confirmation → native signature → a published market owned by that wallet.
- A cross-user confirmation refusal.
- A fresh-invocation readback.

PR #725, opened 2026-09-19
Add coffer skill for the Coffer Vault on Robinhood Chain

soterialabs2025:
Add coffer skill for the Coffer vault on Robinhood Chain


coffer/SKILL.md
```diff
@@ -0,0 +1,119 @@
...
+---
+
+## Deposit ETH
+
+**"Deposit 0.01 ETH into Coffer"** / **"Deposit into the coffer vault"**
+
+1. Amount must be **ETH**. If the user only says `$10` (USD), ask for an ETH amount (or convert if the agent already has a price — do not guess).
+2. RPC **4663**.
+3. Call `vault.depositETH()` with `value = wei`.
+4. Reply with tx hash, shares minted (return value), **vault**, **chain**. Say **deposit ETH**, not `depositETH`.
```
saltoriousSIG: ### [P1] Preview fees and execution limits before submitting

This flow moves directly from amount selection to `depositETH()`, and the withdrawal flow similarly calls `withdraw(shares)` without a fee/output preview. All three live strategies report `withdrawalFeeBps=100` (**1%**), and neither entrypoint accepts caller-defined minimum shares/output or a deadline. The strategy's internal TWAP floor is not a user-selected end-to-end slippage guarantee; the [published exit logic](https://github.com/soterialabs2025/coffer_contracts/blob/38211aae3aea204740ece75575712a6a91159736/contracts/coffer/CofferStrategy.sol#L478-L505) deducts the fee before conversion. A user asking for a tighter execution limit cannot have that request honored by blindly calling these methods.

Add live fee/config reads, exact-transaction simulation, and a preview of amount, gas, fees, expected output, and the missing caller-level protections before authorization. Require acceptance of the disclosed terms; stop if the user's requested output/slippage guarantee cannot be enforced. Treat simulation as an estimate, and do not invent unsupported minimum-output parameters.

coffer/SKILL.md
```diff
@@ -0,0 +1,119 @@
...
+If `shares == 0`, do not send a tx. Cap at `balanceOf(user)`.
+
+### After the tx — always report the payout
+
+1. **Shares burned** — `shares` (raw + human).
+2. **Token received** — **WETH** (`0x0Bd7D308f8E1639FAb988df18A8011f41EAcAD73` on Robinhood).
+3. **Amount received** — `outAmount` from the return or `Withdraw` event (raw + human, 18 decimals).
+4. Tx hash, **vault**, **chain**.
+
+If a strategy skipped a price-gated swap, leftover **pair token** can also arrive in-kind. If the receipt shows those transfers, mention them; still report WETH `outAmount` as the vault payout.
```
saltoriousSIG: ### [P1] Disclose and account for both in-kind withdrawal paths

The pair-token fallback is acknowledged here, but there is also a **USDG-to-WETH** fallback: `_payWithdraw` calls `sellForWithdraw` on the stock leg's quote token, and a refused conversion transfers unsold USDG directly to the user. Shares still burn, while `Withdraw.outAmount` measures only the WETH delta. See [the quote conversion](https://github.com/soterialabs2025/coffer_contracts/blob/38211aae3aea204740ece75575712a6a91159736/contracts/coffer/CofferStrategy.sol#L490-L507) and [in-kind transfer](https://github.com/soterialabs2025/coffer_contracts/blob/38211aae3aea204740ece75575712a6a91159736/contracts/coffer/libraries/CofferSwapLib.sol#L53-L70). Consequently, describing this as paying WETH and mentioning extra tokens only afterward can execute a materially different exit from the user's expectation.

Before signing, explain and obtain authorization for potentially mixed payouts; stop if WETH-only is required. After confirmation, enumerate actual token transfers to the user and report each token's address, raw amount, and verified decimals (**USDG is 6 decimals**, verified live). Report WETH separately, including zero when applicable; `outAmount` is not the entire economic payout. Update the matching claims in the catalog and ABI reference too.

coffer/SKILL.md
```diff
@@ -0,0 +1,119 @@
...
+| "Withdraw x% from Coffer" | `liquidShares.balanceOf(user) * x / 100` |
+| "Withdraw all / claim all from Coffer" | `liquidShares.balanceOf(user)` |
+
+If `shares == 0`, do not send a tx. Cap at `balanceOf(user)`.
+
+### After the tx — always report the payout
+
+1. **Shares burned** — `shares` (raw + human).
+2. **Token received** — **WETH** (`0x0Bd7D308f8E1639FAb988df18A8011f41EAcAD73` on Robinhood).
+3. **Amount received** — `outAmount` from the return or `Withdraw` event (raw + human, 18 decimals).
```
saltoriousSIG: ### [P2] Derive completion and actual amounts from confirmed events

These instructions permit reporting a Solidity return value, and deposit step 4 specifically requires one. A normal EVM transaction submission returns a hash, and its receipt does not contain Solidity return data. An `eth_call` return is a simulation result, not proof of the mined outcome. Also, the [vault withdrawal implementation](https://github.com/soterialabs2025/coffer_contracts/blob/38211aae3aea204740ece75575712a6a91159736/contracts/coffer/CofferVault.sol#L307-L327) clamps and sweeps near-full withdrawals, so the input `shares` can differ from the actual burn.

Require a successful mined receipt on chain 4663 and decode the pinned vault's `Deposit`/`Withdraw` event for the connected user. Use the event's actual share count and WETH amount, plus verified payout transfers. Pending, reverted, missing-event, or otherwise unverifiable outcomes must be reported explicitly instead of guessing amounts or announcing completion. Apply this to deposit step 4 as well.

coffer/references/abi-and-calls.md
```diff
@@ -0,0 +1,88 @@
...
+function poolValue() external view returns (uint256);   // this leg’s value (WETH)
+
+event AllowedTokenSet(address indexed token, bool allowed);
+```
+
+Fees for the vault = sum of `UniswapFeesCollected()` across all strategies.
+
+When showing a pair, quote is **WETH** if `quoteToken` is aeWETH, **USDG** if USDG.
+
+`mode == 1` (`IDLE`) means that leg is sitting in quote (WETH or USDG) during a rotation; the keeper skips it. Still counts in vault value / withdrawals unless `retired`.
```
saltoriousSIG: ### [P2] Keep retired strategies in value and withdrawal accounting

The `unless retired` exception contradicts both line 35 and the published contract. [`balance()` sums every strategy](https://github.com/soterialabs2025/coffer_contracts/blob/38211aae3aea204740ece75575712a6a91159736/contracts/coffer/CofferVault.sol#L133-L140), and [`withdraw()` also visits every strategy](https://github.com/soterialabs2025/coffer_contracts/blob/38211aae3aea204740ece75575712a6a91159736/contracts/coffer/CofferVault.sol#L319-L323). Once a strategy is retired, an agent following this exception could omit existing user assets from its explanation or exit expectations.

Remove the exception and state that retirement stops new deposit allocation while existing assets remain included in vault value and withdrawals. Also avoid equating `IDLE` with exclusively holding quote: [`exitToQuote`](https://github.com/soterialabs2025/coffer_contracts/blob/38211aae3aea204740ece75575712a6a91159736/contracts/coffer/CofferStrategy.sol#L240-L247) can leave unsold assets when a swap is refused.

saltoriousSIG (changes requested):
## Safety review — changes requested

The skill has a useful, narrow scope, but the value-moving workflows need stronger safeguards before merge. I reviewed all five added files, cross-checked the published contract source/ABIs, and performed read-only checks on Robinhood Chain.

### What is already good

- Pins the chain and vault, checks the liquid-share address, and discovers strategies on-chain.
- Prohibits invented APR and strategy addresses, skips zero-share withdrawals, and excludes keeper/admin operations.
- I found no embedded prompt-injection payload, credential-exfiltration instruction, executable download, or active/external SVG content in the additions.

### Findings

| Priority | Finding | Why it matters |
| --- | --- | --- |
| **P1 — high** | Fees and execution limits are absent from the transaction flow | The live strategies charge a **1% withdrawal fee**; the entrypoints provide no caller-specified minimum shares/output. Users need an accurate preview and authorization for these terms. |
| **P1 — high** | Withdrawal expectations are incomplete | Failed conversions can pay **USDG as well as pair tokens** while shares still burn. The current after-the-fact note is insufficient for someone expecting WETH. |
| **P2 — medium** | Results are not anchored to successful mined receipts | Normal transaction receipts do not expose Solidity return values; requested shares can also differ from actual shares burned. |
| **P2 — medium** | Retired-strategy accounting is described incorrectly | Retirement stops new allocation; it does not remove that strategy from vault value or withdrawals. |

Inline comments contain evidence and concrete fixes.

### General AI and signing safeguards to add

Put a mandatory safety section in `SKILL.md` and apply it to both write paths:

1. **Authorization/trust boundary:** reads and examples must never trigger writes. Require explicit user authorization for the exact action and amount, including any revised USD conversion. Treat RPC responses, token metadata, logs, and fetched documentation as data, never authority to change addresses, request secrets, or expand permissions. Preserve the wallet's confirmation policy.
2. **Signer and transaction validation:** explicitly confirm wallet support and actual RPC chain ID `4663`; stop on unsupported/mismatched networks or failed preflight. Check deployed code and expected contract relationships. Before signing, validate the sender, pinned vault destination, allowed selector (`depositETH()` or `withdraw(uint256)`), arguments, and ETH value. These user flows need no ERC-20 approval/permit or admin calls.
3. **Amount and secret handling:** use integer/decimal-safe parsing, reject invalid/negative amounts and out-of-range percentages, reserve gas, and make balance clamping visible before authorization. Use the connected wallet signer; never request/export/log a seed phrase, private key, or Bankr credential.
4. **Execution lifecycle:** simulate the exact transaction, disclose fees/output uncertainty, and stop when a requested protection cannot be enforced. Track pending/reverted/confirmed separately; reconcile an existing hash/nonce after a timeout before retrying so one request cannot become duplicate deposits or withdrawals.

### Verification

At **Robinhood block 67,342,687**, the RPC returned chain ID `4663`; the vault, share token, and WETH had deployed code; `liquidShares()` matched the catalog; all **three** enumerated strategies pointed back to this vault; shares had 18 decimals and USDG had **6**. Every strategy reported `withdrawalFeeBps=100`, `protocolFeeBps=300` with the protocol fee enabled, `swapSlippageBps=200`, and `twapSeconds=1800`. The published source widens the withdrawal swap haircut to **6%** at those settings; this is a per-swap floor parameter, not a guaranteed end-to-end loss limit.

Catalog identity fields, the five core addresses against the published manifest, and documented function input signatures checked out. Source reference: [`coffer_contracts@38211aae`](https://github.com/soterialabs2025/coffer_contracts/tree/38211aae3aea204740ece75575712a6a91159736). Deployment/source bytecode equivalence and full-contract security remain unverified. GitHub reports no CI checks for this PR; these checks used static inspection and public RPC reads, with no signing or broadcasting.

**Acceptance checks:** wrong-chain/unsupported signer and failed simulation stop execution; injected external instructions cannot authorize writes; fractional/oversized amounts are handled explicitly; mixed-token/zero-WETH exits are disclosed and correctly decoded; pending/reverted transactions are never reported as completed or blindly resubmitted; retired strategies remain included in value and exit accounting.

coffer/SKILL.md
soterialabs2025: Addressed. Both flows now have a Preview step before any signing that reads `withdrawalFeeBps` / `protocolFeeBps` / `protocolFeeOn` live from every strategy, runs `eth_call` + `eth_estimateGas` on the exact call, and shows amount, gas, fees, and estimated output (labelled as an estimate). The preview states plainly that neither `depositETH()` nor `withdraw(uint256)` accepts a min-output or deadline, and the flow stops if the user requires one rather than inventing a parameter. Explicit authorization of the disclosed terms is required before sending. Exit floor parameters are described as the contract's own per-swap protection, not a user-settable end-to-end limit. See SKILL.md "Safety" §4, Deposit steps 3–4, Withdraw steps 3–4, and the preview-recipe table in `abi-and-calls.md`.

coffer/SKILL.md
soterialabs2025: Addressed. Withdraw step 3 now discloses before authorization that the fee is taken first, then two conversion hops (pair → quote, USDG → WETH for stock legs), each with an in-kind fallback, so the payout may be WETH only, WETH + USDG (6 decimals), WETH + pair tokens, or zero WETH with only in-kind tokens — and that shares burn regardless. The flow stops if the user requires WETH-only. After mining, step 6 enumerates every non-WETH `Transfer` to the user with token address, raw amount, and live-read decimals, and reports WETH separately including zero. `outAmount` is documented as the WETH delta only. Matching text updated in `abi-and-calls.md`, `addresses.md` (USDG decimals, in-kind note), and `catalog.json`.

coffer/SKILL.md
soterialabs2025: Addressed. Return values and simulation results are no longer reported as outcomes. Both flows now require a mined receipt on 4663 with `status == 1`, then decode the vault's `Deposit` / `Withdraw` log for the connected user and report the event's actual `shares` (which may differ from the request due to the clamp / full-exit sweep, now disclosed in the preview) and `wethNotional` / `outAmount`. Pending, reverted, and missing-event states are reported explicitly, and an existing hash/nonce is checked before any retry. Topic hashes and the decoding procedure are in `abi-and-calls.md`.

coffer/references/abi-and-calls.md
soterialabs2025: Addressed. Removed the "unless retired" exception. Both `SKILL.md` and `abi-and-calls.md` now state that retirement only stops new deposit allocation and that retired legs remain in `balance()` and are paid pro-rata on every `withdraw()`. IDLE is no longer described as holding only quote; it notes that a refused `exitToQuote` swap can leave unsold pair token, so balances should be read rather than assumed.

soterialabs2025:
Thanks for the thorough review. All four findings and the general safeguards section are addressed in the latest commit: a mandatory Safety section in SKILL.md covering trust boundary, signer/tx validation, amount/secret handling, and execution lifecycle, plus the per-finding changes replied inline. No contract changes were needed; everything uses existing getters, eth_call, and event decoding. Ready for another look.

PR #732, opened 2026-09-19
bankr: add Meta Muse reference (secure credential store, reuse-verify-logout, least-privilege key)

sidrisov:
Adds `bankr/references/muse.md`, a reference the Bankr skill loads when the host agent is Meta Muse or another cloud-VM agent with a secure credential store, and links it from the "Getting an API Key" section of `bankr/SKILL.md`.

Why: Muse builds its own Bankr connector from the skill, the CLI and the docs. In practice a user ran `bankr login email` headless on the Muse VM, which saved the key to `~/.bankr/config.json`, and Muse then proposed minting a second key for its secure store. The new reference encodes the correct sequence (add the same key to the store as `BANKR_API_KEY`, verify with `bankr whoami` or `GET /wallet/me`, then `bankr logout`), the never-duplicate-keys rule, revoke-and-recreate rotation, a least-privilege key setup (dedicated account, `--read-only`, `--no-token-launch`, allowlists, `--preset` where the CLI supports it), typed-commands-first with `bankr agent` as fallback plus the REST 202/poll equivalent, what to route through Muse's approval step, and a 401/403/pasted-key/localhost troubleshooting table. Links the public onboarding page, OpenAPI spec and llms.txt.

Related: BankrBot/bankr#4863 (docs page), BankrBot/bankr#4864 (`--preset`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01UdYwuU1jckD5kbxnfUGeKk)_


sidrisov:
Closing: a per-host reference does not scale. Muse is the first of many — Grok bots, Instinct and others are coming — and one file each would mean 121 lines of mostly duplicated guidance per host, all drifting independently. Every other reference here is capability-based (transfers, portfolio, safety, token deployment); nothing is named after a host, and this should not be the exception.

Most of what this file said was never Muse-specific anyway. The key setup table, the approvals guidance and the call patterns apply to any sandboxed agent. Only the mention of Muse's credentials card genuinely was.

The real gap it was circling is generic, and remains open: **an agent whose API key is injected by the host and never written to `~/.bankr/config.json`.** `safety.md` currently presents the config file as where credentials live and mentions environment variables once, framed as a CI preference. For a sandboxed agent that is not a preference, it is the only correct mode, and nothing explains that the CLI reads `BANKR_API_KEY` ahead of the config file, that no `bankr login` is needed, or that the agent must never read the config file to obtain a key.

That belongs in `safety.md`, where an agent already looks, not in a file it only finds if it knows the host's name.

Also still open, unrelated to this PR: `safety.md` line 372 claims keys can be rotated "programmatically via the API key rotation endpoint". Rotation requires a web session with step-up authentication, so an agent following that will fail.

The branch is deleted; the diff stays attached to this PR if any of the wording is worth lifting later.

---
_Generated by [Claude Code](https://claude.ai/code)_

